PR #363, opened 2018-01-09
Use the Fullname in top bar of Web interface and not the login

earendilfr:
To display the display name and not the login on Web interface.
Should be more readable sometime

### DIFF
--- a/lib/App/Netdisco/Web/AuthN.pm
+++ b/lib/App/Netdisco/Web/AuthN.pm
@@ -58,6 +58,7 @@ post '/login' => sub {
 
     if ($success) {
         session logged_in_user => param('username');
+        session logged_in_fullname => schema('netdisco')->resultset('User')->find(param('username'))->get_column('fullname');
         session logged_in_user_realm => $realm;
 
         schema('netdisco')->resultset('UserLog')->create({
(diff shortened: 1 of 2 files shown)


lib/App/Netdisco/Web/AuthN.pm
```diff
@@ -58,6 +58,7 @@ post '/login' => sub {
 
     if ($success) {
         session logged_in_user => param('username');
+        session logged_in_fullname => schema('netdisco')->resultset('User')->find(param('username'))->get_column('fullname');
```
ollyg: Good idea and thanks for the patch! I just want to see whether we can find a cleaner way of getting the fullname than a complete schema call, here, and then I will be happy to merge.

share/views/layouts/main.tt
```diff
@@ -168,7 +168,7 @@
           [% ELSE %]
             <i class="icon-user"></i>
           [% END %]&nbsp;
-            [% session.logged_in_user | html_entity %] <b class="caret"></b>
+            [% session.logged_in_fullname | html_entity %] <b class="caret"></b>
```
ollyg: I think we also want to fall back to username if the full name is not available (I believe full name is optional?).

earendilfr: I will check if I can have a solution for this point :)

ollyg:
Fixed and implemented - thanks!!

PR #374, opened 2018-02-14
Fix two issues

earendilfr:
- Unable to start the netdisco-do renumber function
- Chen we perform check on ACL with negate character, the negate character is deleted during the first match


ollyg:
Hi @earendilfr many thanks for that !! I have released version 2.039004 of Netdisco with the *first* firx related to renumber.
However I am not sure I understand the patch for the second fix. I will need to spend time testing, unless you can provide a test case which passes and fails?

earendilfr:
For the second patch:
I have the following host_groups:
```
host_groups: 
  Site_A:
    - 'op:and'
    - '10.0.0.0/24'
    - '!10.0.0.10-11'
    - '!10.0.0.255'
  Site_B:
    - 'op:and'
    - '10.0.1.0/24'
    - '!10.0.1.10-11'
  Site_C:
    - '10.10.0.0/24'
  Region_AB:
    - 'group:Site_A'
    - 'group:Site_B'
  Region_C:
    - 'group:Site_C'
```

When I pass in the check_acl function, after the first pass, the `!` character is removed..
You can found the result of a printf command inside the check_acl function on my server (Data removed):
```
DEBUG: item=group:xxx | thing=10.23.231.11
DEBUG: item=group:xxx_y | thing=10.23.231.11
DEBUG: item=10.23.230.0/24 | thing=10.23.231.11
DEBUG: item=group:xxx_z | thing=10.23.231.11
DEBUG: item=10.23.231.0/24 | thing=10.23.231.11
DEBUG: item=!10.23.231.12 | thing=10.23.231.11
DEBUG: item=!10.23.231.13 | thing=10.23.231.11
.
.
.
DEBUG: item=group:xxx_y | thing=10.23.231.11
DEBUG: item=10.23.230.0/24 | thing=10.23.231.11
DEBUG: item=group:xxx_z | thing=10.23.231.11
DEBUG: item=10.23.231.0/24 | thing=10.23.231.11
DEBUG: item=10.23.231.12 | thing=10.23.231.11
```

I think the problem is, on the following line:
```
foreach my $item (@$config) {
...
my $neg = ($item =~ s/^!//);
```
it's modified the reference. So the array is definitively modified for future used.

ollyg:
Thank you @earendilfr !
Your explanation is very clear - I have applied the patch and this will be in the next release. Well spotted.
regards,
Oliver.

PR #396, opened 2018-04-12
dump sshcollector stderr into null

bldewolf:
If SSH prints anything to stderr, it leaks out to the caller of sshcollect. This change dumps it into /dev/null instead.

It might be better to collect it and output it for debug purposes, but this cleans up logging in the interim.

I'm only able to test this with the Palo Alto plugin so I apologize if it breaks one of the others.


ollyg:
Many thanks again!

PR #397, opened 2018-04-12
Improve Palo Alto SSH Collector support

bldewolf:
The PAN CLI tries to do friendly auto-completion things and makes the SSH
Collector fail sporadically.  This change uses the "set cli scripting-mode on"
command to calm down the PAN CLI and works around the extra echoed prompts that
get sent.

This change also adds collection of IPv6 neighbor information.

I'm not entirely happy with how I handle the CLI interactions.  Rancid's `panlogin` looks like a good template for how to handle prompts but I got stumped trying to convert tcl expect to perl Expect.

This is tested against PAN-OS 8.  I don't have any other PAN-OS versions to test against, unfortunately.


ollyg:
Many thanks!!

PR #408, opened 2018-04-26
Change on make rancid conf

earendilfr:
Add the following parameters to the MakeRancidConf:

- Change the default delimiter character from ':' to ';' that is the default RANCID delimiter since RANCID 3.x
- Add an option to define the rancid config directory.
  This directory will be used to extract the vendor defined in the RANCID configuration to avoid to extract
   unexisting vendor devices in RANCID configuration (by example, Ubiquiti AirOS devices).
- Add an option to define a default_group.
  If a device doesn't match the defined group, it will add to this group
- Add an option to defined a manually excluded ACL
   To permit to not export devices from Netdisco to RANCID based on an ACL.


ollyg:
Thanks! I amended the patch to make rancid_conf optional, and merged into master at [aefef155..232df964].

Issue #379, opened 2018-03-02
connected nodes all share same LLDP Inventory data


ollyg:
The LLDP inventory data is tied to a port. When that port has multiple nodes such as a phone with pass through switch to a PC, the PC is also displayed as a phone in Netdisco.

We need a way to know that only one neighbor is the phone (or the LLDP inventory source). This has to be done before building reports on phones and waps, otherwise the data will be no good.

ollyg:
With reference to #36

ollyg:
Updated in 2.039013 to no longer display phone or wap icon alongside nodes (unless wireless node info is available)

ollyg:
LLDP-MIB::lldpRemChassisId could work, if ChassisIdSubType = macAddress
and also LLDP-MIB::lldpRemPortId 
and even lldpRemManAddrIfId could report a MAC

Issue #252, opened 2015-10-22
Search with regular expressions and cound of devices matched


netdisco-automation:
Hello,

first thanks for a great tool, it's incredibly useful! I would like to make two feature requests in the "Device Search Options".

To be able to use regular expressions to search devices in the system name and a count of the devices matched.

Kind regards,

Aarón

Reported by: fefa2k

Original Ticket: [netdisco/netdisco2/251](https://sourceforge.net/p/netdisco/netdisco2/251)

Issue #140, opened 2014-04-29
get_port_macs not filtering MACs by device


netdisco-automation:
Hi,

Thx for the great application Netdisco2.   Possible bug: function get_port_macs (in Util::PortMac) returns port MACs for all devices in the database, rather than for just the specified $device.  This prevents function _walk_fwtable from finding uplink ports on older non-CDP/LLDP switches; effectively, macsuck_bleed is always on.
________________________________________
Environment:
$ uname -a
Linux NetDisco 3.13.0-24-generic #46-Ubuntu SMP Thu Apr 10 19:11:08 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

$ perl -v
This is perl 5, version 18, subversion 2 (v5.18.2) built for x86_64-linux-gnu-thread-multi
(with 41 registered patches, see perl -V for more detail)

$ egrep '^\$VERSION' .../perl5/SNMP/Info.pm
$VERSION = '3.13';

$ grep VERSION .../perl5/App/Netdisco.pm
our $VERSION = '2.026000';

Reported by: joelleonhardt

Original Ticket: [netdisco/netdisco2/112](https://sourceforge.net/p/netdisco/netdisco2/112)

PR #409, opened 2018-04-27
Phone icon for nodes

pilyon3:
Removed by issue [#379](https://github.com/netdisco/netdisco/issues/379). Parameter "phone_platforms" doesn't work with all phone types because LLDP field "SysDescr" isn't always available. For these cases, the goal of this modification is to allow the possibility to define a new parameter "phone_mac_vendor" into deployment.yml configuration file in order to list phone's MAC vendor (first three MAC address octets) to match phone nodes MAC address and display the relevant icon.
New parameter "phone_mac_vendor" accepts MAC vendor uppercase/lowercase formats and dash or colon characters.

### DIFF
--- a/lib/App/Netdisco/Worker/Plugin/Discover/Neighbors.pm
+++ b/lib/App/Netdisco/Worker/Plugin/Discover/Neighbors.pm
@@ -138,13 +138,22 @@ sub store_neighbors {
   $c_ip = { %$c_ip, %c_ipv6 };
 
   foreach my $entry (sort (List::MoreUtils::uniq( keys %$c_ip ))) {
-      if (!defined $c_if->{$entry} or !defined $interfaces->{ $c_if->{$entry} }) {
+      my $entry_if = $entry;
+      if ($entry_if =~ /^0\./){
+	       $entry_if =~ s/^0\.//g;
+	       foreach my $temp (keys %$c_if){
+	          if ( $temp =~ /$entry_if/ ){
+              $entry_if = $temp;
+            }
+         }
+      }
+      if (!defined $c_if->{$entry} or !defined $interfaces->{ $c_if->{$entry_if} }) {
           debug sprintf ' [%s] neigh - port for IID:%s not resolved, skipping',
             $device->ip, $entry;
           next;
       }
 
-      my $port = $interfaces->{ $c_if->{$entry} };
+      my $port = $interfaces->{ $c_if->{$entry_if} };
       my $portrow = schema('netdisco')->resultset('DevicePort')
           ->single({ip => $device->ip, port => $port});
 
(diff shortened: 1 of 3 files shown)


ollyg:
We used to have this feature in Netdisco 1, and I never ported it to Netdisco 2 because it seemed unmaintainable to me, and a cause of confusion. So I am not yet ready to push the button on this one, and would welcome other comments. For reference the ND1 settings were called "phone_ouis" and "wap_ouis".

In any case, the patch also needs to set a default for `phone_macvendor` in `share/config.yml`.

ollyg:
(I meant comment in the code, sorry for not being clear - I see there is a commit comment)

@jeneric or @JeroenvIS may wish to add a note on whether this should be done in SNMP::Info c_* methods.

pilyon3:
Hi,

Comments added into code. It could effectively be a good idea to patch SNMP::Info rather than Netdisco.

JeroenvIS:
Agreed, this should be patched in SNMP::Info. Looks like an issue with TimeFilter in the LLDP-MIB implementation on the device, and that would be something to deal with in SNMP::Info::LLDP module itself.

jeneric:
In general I agree that LLDP patch should be addressed in SNMP::Info. There are two issues open in SNMP::Info right now that are probably relevant:

- [#252 Multiple issues in LLDP/CDP discovery](https://github.com/netdisco/snmp-info/issues/252)
- [#140 LLDP fixes](https://github.com/netdisco/snmp-info/issues/140)

I haven't started working on those yet as I was trying to clear out the more straight forward open issues.

I haven't fully dug into the patches here, but the index shouldn't come into play as we should be cross referencing to c_port. For the gear mentioned in this ticket there is work around in SNMP::Info::LLDP in lldp_port(), see '# Avaya/Nortel lldpRemPortDesc doesn't match ifDescr' the regex there may be outdated - also this should be moved to the applicable class rather than in the main LLDP class. The Avaya/Nortel Baystack series was recently acquired by Extreme. Port names are generated (not from MIB) for these devices for compatibility with the proprietary SONMP L2 topology protocol.

pilyon3:
Eventually I've just found an ERS4950 not yet upgraded. This one was in SW version 7.2 and so I've upgraded it step by step from 7.2 to 7.5 version.

The result is that all works fine until 7.4.0.215 which is for now the penultimate version (test with an WAP Aruba as neighbor): 
2018-05-02_13h29_20

Upgrading into 7.5.0.0007 permits to reproduce the problem:
2018-05-02_13h56_11

As I did not find anything that met our case in Extreme's release notes, I think we can consider that's a bug. I'm going to open a case on the Extreme Support Portal.

jeneric:
@pilyon3 thanks for the detailed troubleshooting, as this is a vendor code issue I'm going to close the associated SNMP::Info ticket I created.

ollyg:
Hi all. I am inclined to close this patch. It seems we have a better understanding of the LLDP and can make it work better, and I'm still not comfortable with the MAC Vendor based matching for phones and WAPs (fragile and high maintenance).

Please comment if you believe we ought to have this feature still.

PR #417, opened 2018-05-14
Change on make rancid conf

earendilfr:
Hello,

I have review the change after the PR #408.
It seems better (and I have take your modification from your commit.

I have tested inside my personnal instance and no error after a simple run of netdisco-do makerancidconf


earendilfr:
Please don't take in count currently, I still have an issue with the vendormap...

earendilfr:
OK, Finally found: I still have the old format for the vendormap.
I have also added a small typo to sort the entry in router.db file

From my point of view, the PR is OK.

ollyg:
Hi @earendilfr I am a bit confused by this PR now as it seems to revert some of your own changes. I already committed most of your suggestions to  the released version of MakeRancidConf.

Please would you open a new pull request if there are any outstanding issues with the current release of Netdisco? Thanks.

PR #427, opened 2018-07-13
update NodeVendor.pm

slofunk:
It would seem like we'd want for this to be distinct mac addresses here right? Otherwise we'd could get duplicate counters. In our environment where things move around a bit this makes for repeat macs in the nodes table. To get a true inventory count I'd think this should only count unique macs. Apologies if this is NOT the desired output for this module.


ollyg:
thanks!!

PR #438, opened 2018-09-05
Update config.yml

inphobia:
these are used internally in cisco ucs blade chassis, there is no reason i can think of to collect these.


inphobia:
side note:

in the list of ignored interfaces you also have:
`- 'M-GigabitEthernet\d/\d/\d'`

any reason why the forward slashes are not escaped?

ollyg:
thanks!

PR #459, opened 2018-11-28
1 more isdn interface to ignore

inphobia:
not sure if this is a typo or not, but bri singaling interfaces have 1 'l' on cisco devices.

"BRI0-Signaling" was taken from a netdisco report, router is a "897VaK9" running ios "15.5(3)M7".

as such, i've added the 1 'l' variant to the ignore_interfaces list. if this was indeed a typo perhaps the double 'll' variant can be deleted.


JeroenvIS:
Not sure if it was a typo, would be hard to check. Doesn't hurt to have both in the default config. Thanks for the PR.

PR #460, opened 2018-12-06
Improve netdisco-sshcollector CLI experience

rc9000:
Hi netdisco team! I was debugging some sshcollector failures on IOSXR. It was a bit cumbersome with the current code, so I added some additional command line options. I hope you find this useful too.


inphobia:
works nicely for me, but it would be nice if there could be a more prominent error on connection failures. now i need to activate -O to see:
# set_error(1 - unable to establish master SSH connection: password authentication failed)
the other error is just a:
[19234] 2018-12-08 21:57:51  warn WARNING: no entries received from ...

side note: can this be run from a netdisco backend worker or is it cron only?

thx

rc9000:
Thanks for checking it out @inphobia 

Yes good point, in fact the poor error reporting has mislead me before. I've added a check for a successful Net::OpenSSH connection, and a warning on failure. Now we get e.g. 

    bin $ netdisco-sshcollector -d device_with_nonexistent_hostname 
    [388] 2018-12-10 22:24:51  warn WARNING: Couldn't connect to <device_with_nonexistent_hostname> - unable to establish master SSH connection: bad password or master process exited unexpectedly
    bin $ netdisco-sshcollector - device_with_wrong_password 
    [2115] 2018-12-10 22:25:37  warn WARNING: Couldn't connect to <device_with_wrong_password> - unable to establish master SSH connection: password authentication failed

The other new commit just fixes the POD and Getopt handling for the -Q flag, in fact it doesn't look like there is any point in specifying it multiple times.

rc9000:
Regarding the cron question: No, currently netdisco-sshcollector can only be run from cron. But I think it would be nice if it would run from the daemon like the normal Arpnip job.  IMHO, if there is an entry for a device in deployment.yml->sshcollector, the job could just run the sshcollector code instead of the SNMP-based arpnip.

Maybe if the @ollyg, @JeroenvIS  etc. are on board and agree with the spec,  I could add this as a separate PR. In our network with a lot of BigIP and VRF-heavy IOSXR/NXOS, SNMP-based arpnip is pretty much a thing of the past.

inphobia:
coolness, once you're done with this i might expand it accept vrfs as a parameter. on some devices we have cli access but not the required authorization to show all vrfs, just select a select few. trying to show all either show no arp entries or an authorization error on those.
or perhaps to leave out, in the nx-os case, the vpc keepalive or management vrf...

ollyg:
Thanks!

ollyg:
Hi @rc9000, yes please go ahead, this would be very welcome. It may be less trouble than you think after some recent additions to Netdisco's core. Off the top of my head:

1. Copy https://github.com/netdisco/netdisco/blob/master/lib/App/Netdisco/Worker/Plugin/Arpnip/Nodes.pm and rename to something new such as Arpnip/NodesBySSH.pm
1. Within that file... after getting `$device`, delete the SNMP bit, it's not needed, and instead find the ssh collector config entry in settings and retrieve user/pass etc
1. Obviously return Status->info('nothing to retrieve') if there's no config for the device
1. Spawn the SSH client
1. Load and run the ssh collector as in netdisco-sshcollector
1. Store the entries using store_arp (copy from netdisco-sshcollector)

You may wish to read https://github.com/netdisco/netdisco/wiki/Backend-Plugins#worker-return-status and https://github.com/netdisco/netdisco/wiki/Worker-Cookbook to see more about the features available to Workers.

rc9000:
Ok great, thanks for the hints. I'll get on it, no promise on ETA for now though :)

inphobia:
side note: is this still correct?
"If the password is "-", public key authentication will be attempted."

i need to have a blank password if i want pubkey to work.

rc9000:
@inphobia 
> "If the password is "-", public key authentication will be attempted."
> i need to have a blank password if i want pubkey to work.

Yes I think you're right, it's probably wrongly documented like that since the Netdisco 1.3 version of the script. I've created a new issue #473 for this.

I think we should also allow passing passphrase and key_path options from the deployment.yml->sshcollector config to make this useful for the general case.

Issue #461, opened 2018-12-06
SSHCollector::Platform::IOSXR not working when running from cron (without stdin)


rc9000:
## Expected Behavior
netdisco-sshcollector should collect ARP entries from IOSXR when run in a terminal or from cron.

## Current Behavior
We noticed that some ARP entries from our ASR were not collected. But to our surprise, when running the netdisco-sshcollector in a normal shell, it worked just fine.

After some digging in the ssh -vvv output, we discovered the following difference:

```text
ssh DEBUG3 from terminal, with stdin available

debug2: process_mux_new_session: channel 1: request tty 0, X 0, agent 0, subsys 0, term "xterm", cmd "show arp vrf all", env 2
debug3: process_mux_new_session: got fds stdin 6, stdout 7, stderr 8
debug2: fd 7 setting O_NONBLOCK
debug1: channel 2: new [client-session]
...
debug2: channel 2: open confirm rwindow 2097152 rmax 32768
debug1: mux_client_request_session: master session id: 2
debug3: receive packet: type 99
debug2: channel_input_status_confirm: type 99 id 2
debug2: exec request accepted on channel 2
# stdout, bytes read: 4 at offset 0  
```

```text
ssh DEBUG3 from cron, no stdin available

debug2: process_mux_new_session: channel 1: request tty 0, X 0, agent 0, subsys 0, term "xterm", cmd "show arp vrf all", env 2
debug3: process_mux_new_session: got fds stdin 6, stdout 7, stderr 8
debug2: fd 6 setting O_NONBLOCK
debug2: fd 7 setting O_NONBLOCK
debug2: fd 8 setting O_NONBLOCK
debug1: channel 2: new [client-session]
...
debug2: channel 2: open confirm rwindow 2097152 rmax 32768
debug2: channel 2: read<=0 rfd 6 len 0 <-- read on fd6
debug2: channel 2: read failed  <-- show command fails here, and sshcollector returns no results
```

## Possible Solution
So we found a first solution, which is to just run it from cron like:

    20 * * * * netdisco-sshcollector < /dev/zero 

A nicer solution inside the module:

```text
diff --git a/lib/App/Netdisco/SSHCollector/Platform/IOSXR.pm b/lib/App/Netdisco/SSHCollector/Platform/IOSXR.pm
index 37d44e00..a3b1210c 100644
--- a/lib/App/Netdisco/SSHCollector/Platform/IOSXR.pm
[...]

PR #462, opened 2018-12-06
Potential fix for #461

rc9000:
Please see the comments in the issue for details. In short, IOSXR show commands seem to require an open stdin to be passed from Net::OpenSSH?


ollyg:
Thanks!

PR #463, opened 2018-12-06
In discovery, ignore 'notPresent' interfaces

postilion:
During discovery, ignore all interfaces of type 'ethernetCsmacd'
which are described as 'notPresent'.

The CiscoSB class devices will pre-define phantom interface place-
holders in the MIB for any physical interface which /might/ be in
the stack.  In a large stack this may lead to hundreds of such
phantom interfaces, which just take up vertical space in the GUI
and rows in the DB.

By checking for type == ethernetCsmacd, we allow the Port-Channel
and Tunnel interfaces, which also get described as notPresent, to
appear.

### DIFF
--- a/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
+++ b/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
@@ -181,6 +181,13 @@ register_worker({ phase => 'early', driver => 'snmp' }, sub {
           next;
       }
 
+      # Skip physical interfaces which are physically 'notPresent'
+      if (defined $i_up->{$entry} and $i_up->{$entry} eq 'notPresent' and $i_type->{$entry} eq 'ethernetCsmacd') {
+          debug sprintf ' [%s] interfaces - ignoring %s (%s) (%s)',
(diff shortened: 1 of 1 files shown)


postilion:
@ollyg @netdisco/netdisco-developers 
Nick Nauwelaerts suggests adding a new config item,  "'ignore_nonpresent_types', since removing tunnel & portchannels seems tempting. also, this way if you dont want to ignore nonpresent interfaces you can define a blank list."  I'm inclined to agree with this, but since I am new to Dancer apps, it's not really clear to me how to proceed with that task.

I've looked over lib/App/Netdisco/Configuration.pm, and lib/auto/App/Netdisco/config.yaml, and that much looks straight forward, but before I start hacking away at it, could someone either bless this or warn me off?

Thanks!

JeroenvIS:
@postilion Good suggestion to skip these interfaces. And even better to have the check configurable, so that it's possible to skip it completely (empty type list) or also apply it to other interface types, like tunnels.

Looks like @inphobia already took care of your concern on how to proceed in #464 ;-)

ollyg:
Dupe of #464

PR #464, opened 2018-12-08
ignore notpresent interfaces, with type filtering config item

inphobia:
as #463 by nic. but now with config item to select types to filter, which can be set to [] to not filter at all.

went through my device database, seems only cisco small business switches have "notpresent" interfaces at first glance. this can be either down/notPresent or up/notPresent.

side note: really old ros versions (1.1.2.0) report port channel interfaces as ethernetCsmacd.

### DIFF
--- a/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
+++ b/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
@@ -181,6 +181,13 @@ register_worker({ phase => 'early', driver => 'snmp' }, sub {
           next;
       }
 
+      # Skip interfaces which are 'notPresent' and match the notpresent type filter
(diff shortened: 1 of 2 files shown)


JeroenvIS:
Agree on the approach. I believe only the documentation update is not included yet; if that's added and all tests pass, I'd say its good to go.

inphobia:
correct, no documentation included yet. i used 'ignore_interfaces' as my template, which only seems to be defined in:
config.yml & Discover/Properties.pm
according to a code search:
https://github.com/netdisco/netdisco/search?q=ignore_interfaces&unscoped_q=ignore_interfaces

i did plan to update the wiki page should this code be commited, but only then as to not cause confusion.

if you can point me to where else docs are expected i'll add a blurb for both this settings & ignore_interfaces there (since it was not quite clear to at first if this was case sensitive or insensitive, greedy or exact matching). i seem to recall reading a perl 3pm manual a few years ago but the file referenced in config.yml:
"See App::Netdisco::Manual::Configuration for explanations"
seems to be missing in my installs & src checkouts?

ollyg:
Thanks! I'll update the docs now

PR #465, opened 2018-12-12
workaround for #457

### LINKED ISSUE #457: device address sorting not working for "interface" field
device address sorting not working for "interface" field, but only on specific devices.

## Expected Behavior

when clicking on a device, then going to "addresses" and the clicking on the "interface" table header, netdisco should sort the interfaces.

## Current Behavior

when doing this on a cisco nexus 9332pq switch for some reason my browser doesnt sort & just shows "processing" when using the interface table header to sort. other table headers like address or dns work fine, and on other devices "interface" also sorts as expected.
citrix netscalers also seem to have this issue.

## Steps to Reproduce (for bugs)

1. have access to a nexus switch or netscaler load balancer
1. click on device in netdisco
1. go to addresses
1. sort on interfaces by clicking on the table header
1. notice "processing" overlay in browser but no sorting happens
1. other headers sort just fine

i've gone over other device types as well but only netscaler & nx-os seem to have this issue. when trying to find a correlation i came up with the following:

for both the netscalers & nexus 9332s i notice:
* we have multiple addresses
* all ip addresses are shown correctly ("address" field)
* for netscaler all prefixes are shown, for nexus 1 prefix is missing
* at least 2 ports do not have dns entries (which is correct since some addresses dont have reverse dns)
* at least 2 ports dont have dont have their interface filled in. for the nexus these are ports that belong to different vrf's, netscalers dont have any interfaces filled in.
* the lack of interface names for __at least 2 ports__ seems to be the key here, we also have fortinet devices where only 1 interface is not filled in, and these sort correctly.

so, you need a device that has multiple interfaces, where at least 2 interfaces do not have an interface name and also lack dns entries (perhaps description too), then sorting will fail.

image

## Your Environment

* Netdisco version used: 2.39.33 & 2.39.31
* SNMP::Info version used: 3.63 & 3.61

inphobia:
i'm not in anyway versed in javascript, but this hack fixes #457 for me.

### DIFF
--- a/share/public/javascripts/portsort.js
+++ b/share/public/javascripts/portsort.js
@@ -30,6 +30,8 @@ function portSort (a, b) {
         // octal regex
         ore = /^0/,
         // convert all to strings and trim()
+        if (typeof a == 'null') { a = ''; }
+        if (typeof b == 'null') { b = ''; }
         x = a.toString().replace(sre, '') || '',
         y = b.toString().replace(sre, '') || '';
 


ollyg:
Thanks, I will fix the tests later

Issue #466, opened 2018-12-13
discover jobs are never dequeued or skipped for non discoverable devices


inphobia:
## Expected Behavior

devices should be removed from the jobqueue after they reach max_deferrals or placed on hold somehow.

## Current Behavior

devices which we cannot connect to via snmp remain in the jobqueue it seems. they also seem to be enqueued multiple times.

## Steps to Reproduce (for bugs)

discover job for this device is queued multiple times:
```
netdisco-do psql -e "SELECT * FROM public.admin WHERE device = '10.114.1.14'"
[28994] 2018-12-13 19:19:46  info App::Netdisco version 2.039033 loaded.
[28994] 2018-12-13 19:19:46  info psql:  started at Thu Dec 13 20:19:46 2018
  job   |          entered           | started | finished |   device    | port |  action  | subaction  | status | username | userip |         log         | debug |    device_key
--------+----------------------------+---------+----------+-------------+------+----------+------------+--------+----------+--------+---------------------+-------+-------------------
 107769 | 2018-12-12 23:09:03.581714 |         |          | 10.114.1.14 |      | discover | with-nodes | error  |          |        | duplicate of 102815 |       | ec:74:ba:18:f5:7c
 105298 | 2018-12-12 19:08:35.60186  |         |          | 10.114.1.14 |      | discover | with-nodes | error  |          |        | duplicate of 107769 |       | ec:74:ba:18:f5:7c
 102815 | 2018-12-12 15:06:21.367012 |         |          | 10.114.1.14 |      | discover | with-nodes | error  |          |        | duplicate of 107769 |       | ec:74:ba:18:f5:7c
 110232 | 2018-12-13 03:06:24.198368 |         |          | 10.114.1.14 |      | discover | with-nodes | queued |          |        |                     |       | ec:74:ba:18:f5:7c
 112708 | 2018-12-13 07:08:44.263661 |         |          | 10.114.1.14 |      | discover | with-nodes | queued |          |        |                     |       | ec:74:ba:18:f5:7c
 115186 | 2018-12-13 11:13:28.062473 |         |          | 10.114.1.14 |      | discover | with-nodes | queued |          |        |                     |       | ec:74:ba:18:f5:7c
 117659 | 2018-12-13 15:11:36.075763 |         |          | 10.114.1.14 |      | discover | with-nodes | queued |          |        |                     |       | ec:74:ba:18:f5:7c
[...]

PR #469, opened 2018-12-14
make job queue list in webinterface configurable

inphobia:
related to #466 

add an option to set the amount of jobs returned by the job queue webinterface.  when a discoverall fires off we have around 1000 jobs in the queue. the job queue webinterface does not always seem to show running or queued jobs before completed ones.

with the default of 50 sometimes we just get a list of completed jobs with no way to check (or stop/delete) if there are queued or running jobs. this adds a option in the configuration so you can increase the amount of jobs shown. i prefer 200 jobs with a refresh rate of 30 seconds.

perhaps a dropdown box based on table_showrecordsmenu would also be an option.

or a sorting preference (first running, then queued, then error, then done, etc), but then you will also need to define a maximum timeframe based on job submission. but that's overengineering it imo.

### DIFF
--- a/lib/App/Netdisco/JobQueue.pm
+++ b/lib/App/Netdisco/JobQueue.pm
@@ -59,8 +59,9 @@ Performs initialisation of the Job Queue backend.
 
 =head2 jq_log()
 
-Returns a list of the most recent 50 jobs in the queue. Jobs are returned as
-objects which implement the Netdisco job instance interface (see below).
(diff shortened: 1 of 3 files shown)


inphobia:
docs can be found here: inphobia/netdisco-wiki#2

ollyg:
Thanks, and I will update config guide now...

ollyg:
https://github.com/netdisco/netdisco/wiki/Configuration#code-jobs_qdepth-code

PR #470, opened 2018-12-20
docs are maintained in wiki

inphobia:
per @ollyg:
"the manual files were all converted into GitHub wiki pages"


ollyg:
Thanks!!

PR #483, opened 2019-01-08
add a new tab to device -> vlans

inphobia:
this adds a new tab to device details: vlans

it shows all the vlans know for the device. you can click the vlan id to go to the device ports overview & see all ports with that vlan tagged or native.
clicking on the description goes to reports->vlans

devices with no known vlans give a "no matching records"

csv export also included

preview:
image


inphobia:
my first try at a new plugin. i think i've tested everything but would like some extra eyes.

ollyg:
My only question is whether it should be Vlans or VLANs in the interface :-D

inphobia:
if you ask me, it should be "vlans". caps are such a waste of 26 bits :)

inphobia:
according to the source all lowercase is the clear winner :)

```
grep -r VLAN . | wc -l
134
grep -r Vlan . | wc -l
72
grep -r vlan . | wc -l
481
```

but it seems VLAN is the clear winner in everything that's user output, the only other use of "Vlan" is in search/ports. i'll update them to VLANs to make things uniform.

PR #489, opened 2019-01-17
Integrate netdisco-sshcollector into Worker::Plugin architecture

rc9000:
As laid out in  https://github.com/netdisco/netdisco/pull/460#issuecomment-450528805, this is a first attempt at integrating netdisco-sshcollector into the daemon as plugin.

However there are about as many questions as lines of code, so for now this is a request for comments rather than a real PR. Below are the most important TBD items from the module:

1. Registering a `cli` worker  kinda comes with the requirement to have the credentials defined in `setting(device_auth)`. So my instinct is to abandon the whole `setting(sshcollector)` in favour of it. Surprisingly, there is already some talk about sshcollector-tagged entries in https://github.com/netdisco/netdisco/wiki/Configuration#code-device_auth-code, even though I don't think they would have any impact in the released code. Am I on the right track here or would you prefer to keep `setting(sshcollector)` around?
1. Should we continue running snmp-based arpnip even when ssh-based is configured? If not, how can NodesBySSH.pm block Nodes.pm execution? The https://github.com/netdisco/netdisco/wiki/Backend-Plugins instructions seem to depend on using the same name - which we can't have since it's in the core package and not NetdiscoX.
1. Can we just remove the old bin/netdisco-sshcollector and the sshcollector setting from one release to the other, or should we have a deprecation phase in which both ways work?

There are more questions in the code but I think these is the elementary stuff that should be discussed before we go much further.

Cheers
Christian


rc9000:
Thanks for the reply @ollyg! I have added a very minimal App::Netdisco::Transport::CLI. It does only somewhat align with the SNMP counterpart:

- no distinction of read/write in SSH, so I called the main function `session_for()`.
- caching established connections seems error prone since the caller might close them, network issues etc.
- we have no place to store the successfully used user/pw as in device.snmp_comm etc, but I'm not even sure if sending a list of credentials to a potentially untrusted device is a good idea in the first place.

So I currently just use the first device_auth entry with tag:sshcollector and driver:cli, and return a fresh connection every time.

I also tried to come up with the ACL matching that `Transport::CLI::session_for` should do, but I'm not quite sure how to use check_acl and friends correctly to get the best of potentially a bunch of matching device_auths. Could you sketch out how a sophisticated solution for that would look like? Or should we just stick to an exact match on e.g. device_auth.ip as the old sshcollector setting did? The many additional attributes like the platform etc. might make group-matching on e.g. subnet less useful than it is in SNMP.

Cheers
Christian

inphobia:
i've been wanting to have a look at this for the past few days but other things got in the way.

> > 1. Registering a `cli` worker  kinda comes with the requirement to have the credentials defined in `setting(device_auth)`. So my instinct is to abandon the whole `setting(sshcollector)` in favour of it.
> 
> I agree with this approach. It's a bit embarrassing why I got stuck... I think we already have a "username" field in device_auth for SNMP and I couldn't work out how to distinguish between device_auth entries for SNMP and for SSH if they both have this field (as the "password" field is optional for SSH with a key?). Perhaps I completely missed a really easy solution, or am trying to be too smart about it.

i thought there was a real easy solution, since net::openssh uses the openssh binary as it's backend then why not use the openssh config either under $HOME/.ssh or on some user defined location. this ofcourse has the huge drawback of not actually integrating it into the netdisco config, on the other hand it's a format users might at least be familiar with. but after reading through net::openssh source it seems it's unclear to me which parts, if any, of ssh_config it uses.

there are 2 valid alternatives perhaps: net::ssh:perl which seems to support & use most of the ssh config files, but doesn't use the openssh binary (full perl code it seems). after looking a bit more i found a module that does seem to use to openssh binaries & config: net::cli::interact (with a familiar author 😄 )

that were all the drawbacks, now the major benefits of using the openssh binaries & config:
 *  no dependencies between netdisco & openssh versions. if options are changed in non backwards compatible ways, or ciphers get dropped, or whatever else you can think of, we will not need to add code to handle different versions.
 * seems like an easy way to handle ssh keys instead of passwords too. else we would need to figure out where to store those keyfiles (parsed in netdisco config yml, files on filesystem, ...)
* no need to take into account all possible options, or am i the only one who has devices that run old ssh version, so i need things like this is my .ssh/config:
```
    Host 1.2.3.4
      User RW
      KexAlgorithms +diffie-hellman-group1-sha1
      HostkeyAlgorithms ssh-dss
```

using this setup also takes care of:

>   we have no place to store the successfully used user/pw as in device.snmp_comm etc, but I'm not even sure if sending a list of credentials to a potentially untrusted device is a good idea in the first place.

also, i would like to propose to make host key checking the default. if a device's key changes we should fail instead of just assuming it's normal.


a completely different but perhaps also valid way of looking at it: what we basicly want to do is log into a host, run some commands & parse the output. but then you need to handle different prompts, perhaps log in & then change user, etc... so why bother with trying to do the whole login part when we can just use rancid for that ( http://rancid.shrubbery.net/rancid/svn/rancid/trunk/rancid/README - support a whole lot of device prompts and subtle device bugs ). would open up a whole range of devices for cli interaction, but might also be quite complex to get going.

rc9000:
Hi @inphobia 

> after reading through net::openssh source it seems it's unclear to me which parts, if any, of ssh_config it uses

I think it uses all of them that are not explicitly overridden with options passed to `Net::OpenSSH->new()`. It's not in the Perl code since the ssh binary is doing all of this. 

> huge drawback of not actually integrating it into the netdisco config

Yes I agree, especially since half of the config will still be in deployment.yml (you can't pass a password in ~/.ssh/config, nor the Perl module we need to use for the specific device).

> valid alternatives 

Yeah if anything, we should go to something more high-level like rancid or Net::Interact::CLI. I have never used rancid though, can it handle logic like "log in, show all vrfs, execute a command for each of these"? What about maybe [oxidized](https://github.com/ytti/oxidized) which seems to be a more modern rancid with an API? 

Anyhow this is a pretty big makeover which will require rewriting and testing all the device-specific modules, to some of which none of us might have access too. Sounds quite ambitious for my first substantial netdisco commit :) So I'm leaning to only integrating the current functionality into the daemon.

I think you raised good points though regarding keys, StrictHostKeyChecking etc. The YML should have an item that just allows to pass an array of SSH options, so you can set IdentityFile, HostKeyAlgorithms etc. And as you say, the out-of-the box code should be secure by default, so I'm in favour of axing  StrictHostKeyChecking=no.

Cheers
Christian

inphobia:
> I think it uses all of them that are not explicitly overridden with options passed to `Net::OpenSSH->new()`. It's not in the Perl code since the ssh binary is doing all of this.

after having a second look at the net::openssh code that seems to be the case indeed, and options not directly supported can still be passed with https://metacpan.org/pod/Net::OpenSSH#ssh_opts-=%3E-\@opts . 

> > huge drawback of not actually integrating it into the netdisco config
> 
> Yes I agree, especially since half of the config will still be in deployment.yml (you can't pass a password in ~/.ssh/config, nor the Perl module we need to use for the specific device).

passwords will indeed need to be stored in deployment.yml, and a way to open password protected private keys too. ssh-add, ssh-agent & friends can also be used for this, but net::openssh says it will not run it for you.
i'm not convinced you need to provide a device type ssh perl module. if we're running arpnip on a device, the device has been discovered (don't know of any way to run arpnip on an undiscovered device from within netdisco, sshcollector updated the netdisco db directly without passing through the regular api's & safeguards iirc). and if it's discovered we should already have the info needed to figure out what module to use based on the device type snmp::info returns.

the thing i would like to avoid most however is having the same option defined in multiple places. that's why i'm more in favor of using deployment.yml for the things ssh_config doesn't accept and having the rest in ssh_config.
also, if it indeed does use the defaults values which are not explicitally overridden, we are using the compiled in defaults, then the /etc/ssh client config & the user's ssh_config already anyway. if we can't avoid using those, why not go all in instead? 

it might also make troubleshooting easier: if ssh works on commandline, it should also work from netdisco.

one point of attention in any case will be matching the device name in deployment.yml with the host directive in ssh_config. host and host.example.com are not always equals, and will create different entries in known_hosts too. wildcare matching will also be a point of attention.

> > valid alternatives
> 
> Yeah if anything, we should go to something more high-level like rancid or Net::Interact::CLI. I have never used rancid though, can it handle logic like "log in, show all vrfs, execute a command for each of these"? What about maybe [oxidized](https://github.com/ytti/oxidized) which seems to be a more modern rancid with an API?

a major difference here is that rancid uses interactive mode while net::openssh uses batch mode since it runs an ssh master. rancid's login scripts can be used in a lot of ways:
 * without options it will log you in and set up a terminal for interactive usage.
 * a command or static list of commands can be given as argement.
 * a static list of commands can be read from a file
 * for complex operations it can also use expect scripts.
 * it will also handle additional password prompts, like when you need to run enable or change user in some way to have access to certain commands (like in more complex tacacs setup for example).

it's most common use it to make backups of network devices, where the process is something like:
 * run device login script to log in and set up environment to handle device specifics, like pager prompts.
 * run list of commands specified in it's os config
 * pipe all of that to a perl module for post processing
 * commit it to a revision control system

oxidized is a very valid alternative, i'm just more familiar with rancid having used it for almost 10years now i think. they both have their pro's and cons. if you have netdisco installed however you will already have most of the requirements for running rancid covered, since it needs openssh, perl & expect. oxidized is a ruby application. another point for rancid is that we already have a link with it, since we can generate rancid config files.

> Anyhow this is a pretty big makeover which will require rewriting and testing all the device-specific modules, to some of which none of us might have access too. Sounds quite ambitious for my first substantial netdisco commit :) So I'm leaning to only integrating the current functionality into the daemon as a first step.

totally agree. but everything related to both ssh and cli interaction get's way to complicated very quick. using an external project for all of this has a lot of drawbacks, but the alternative of writing and maintaining it ourselves is also a timesink, more so if we need console interaction for more as basic show commands.

> I think you raised good points regarding keys, StrictHostKeyChecking etc. The YML should have an item that just allows to pass an array of SSH options, so you can set IdentityFile, HostKeyAlgorithms etc. And as you say, the out-of-the box code should be secure by default, so I'm in favour of axing StrictHostKeyChecking=no.

that will need https://metacpan.org/pod/Net::OpenSSH#ssh_opts-=%3E-\@opts for a decent amount of options that are not directly supported in net::openssh.


i didn't mean to make it more sound more complicated as it is, i'm really looking forward to having the sshcollector integrated. my current needs are just being able to use non password protected keys, not ignoring changed keys by default would be a nice bonus. all the rest of this comment could just as well be moved to wishlist or the netdisco3 idea box :)

rc9000:
Ok @ollyg  and @inphobia , this is a first version that I could imagine merging. Sorry it does not change all the backends to Rancid :). But it ports all the functionality into the Job and Transport::CLI.

In this version it's possible to supply master_opts to Net::OpenSSH, e.g. 

    - tag: sshcollector
      driver: cli
      platform: IOS
      username: netdisco
      ssh_master_opts: 
          - "-o"
          - "StrictHostKeyChecking=no"
          - "-i"
          - "my_id_rsa" 

A noteable limitation: if there are multiple matching device_auth blocks, the first will be tried and that's it.

For the migration/deprecation of the standalone script, I have added a big warning to it:

    [19895] 2019-02-01 00:13:33  warn DEPRECATION WARNING
    This script and the sshcollector setting will be removed in a future release!
    See this document to migrate to the new sshcollector integrated into
    regular netdisco-do/netdisco-daemon arpnip:
    https://github.com/netdisco/netdisco/wiki/bin-sshcollector-deprecation

So we can just let it sit in the distro for a couple of releases, and then remove it. 

Let me know if you generally agree with the code, then I will also update all the documentation before we merge.

I will now also test it for a while in our lab to see if there's any negative impact on the daemon.

Cheers
Christian

inphobia:
coolness. regretfully i won't have time in until mid next week to give a a decent test.

the rancid thingy was just an idea, i'm not bound to it in any way (other as i'm also using it and have written the aerohive support for it), it was more of a reflection on the way we want to go forward.

i don't think the device_auth limit you mention is that big of a deal, if you look at https://github.com/netdisco/netdisco/wiki/Configuration#device_auth is also states:

> When configuring multiple SNMPv3 stanza please use only and/or no ACLs for each, otherwise only the first stanza is ever used (this is a limitation in the underlying SNMP library).

while i never ran into that issue (for me it seems to try all snmpv3 options), imo it's a valid option to have first match rule on ssh auth.


testing & code review will regretfully have to wait a few days for me....

rc9000:
Not an expert but seems to be mostly a non-issue today, even the US has realized it's not enforceable: https://en.wikipedia.org/wiki/Export_of_cryptography_from_the_United_States#Current_status

Also, technically we only require Net::OpenSSH which contains no crypto since it just wraps the binary. And the binary is only required the moment somebody adds a CLI transport entry to device_auth. We probably ship more crypto in Web and SNMPv3 code than we do here.  

So whatever the murky status of "exporting" multinational-Netdisco via Github to random downloaders worldwide is, it probably doesn't get much murkier.

inphobia:
from that article:

>  and even cryptographic consulting services still require an export license

while the whole thing is to much legal speak for me, i'm reading that that either americans or everone from wassenaar agreement participants can't help out with setting up ssh or snmpv3 w/ crypto. i guess it has become more efficient to try & backdoor devices (the netscreen/juniper one was quite impressive imo) or algorithms as try & enforce export rules.

rc9000:
Thanks @ollyg for the review, I'll try to do it this way and report back. I think it makes sense that the CLI worker can replace the SNMP variant. 

Also I didn't like that Arpnip::Nodes was for transport SNMP and Arpnip::NodesBySSH was for transport CLI, I was already thinking about renaming everything to Arpnip::Nodes$TRANSPORT to be more coherent - having everything in Arpnip::Nodes solves that problem too.

rc9000:
I have merged NodesBySSH.pm into the regular Nodes.pm, it works exactly as @ollyg promised :)

This will now by default skip the SNMP arpnip if the CLI arpnip was successful. However I've also added an option to still run it if desired.

I have also updated the doc in https://github.com/netdisco/netdisco/wiki/Configuration#code-device_auth-code with all the options available to `tag:sshcollector`. 

In our lab, the daemon seems to run fine since a few days while running SSH workers at a high frequency. If there are no more concerns, I think we should be good to merge.

What do you guys think, ok for you like that?

inphobia:
side note: will this still work with pubkey auth? if so i'll try to get it set up on my test setup & let it run for a bit as well.

thx

rc9000:
@inphobia yes it should work as before. If your key has a standard name/location (.ssh/id_[rd]sa) it will work without any config option. For other keys, you can now pass -i $key in ssh_master_opts. If the password is empty or not should not matter anymore, but you'll need public key before password in the PreferredAuthentications option - this is a default though.

bin/netdisco-sshcollector
```diff
@@ -176,6 +180,15 @@ sub store_arpentries {
     }
 }
 
+sub deprecation_warning {
+    print "\n";
+    warning sprintf "DEPRECATION WARNING\n" .
+                    "This script and the sshcollector setting will be removed in a future release!\n".
+                    "See this document to migrate to the new sshcollector integrated into\n" .
+                    "regular netdisco-do/netdisco-daemon arpnip:\n" .
```
inphobia:
```suggestion
                    "regular netdisco-do/netdisco-backend arpnip:\n" .
```

lib/App/Netdisco/Worker/Plugin/Arpnip/Nodes.pm
```diff
@@ -63,4 +66,78 @@ sub get_arps {
...
+    $device->ip, $arpentry->{mac}, $arpentry->{ip}, 
+    $arpentry->{dns} if defined $arpentry->{dns};
+    store_arp({
+        node => $arpentry->{mac},
+        ip => $arpentry->{ip},
+        dns => $arpentry->{dns},
+      });
+  }
+
+  return 1;
```
inphobia: is that return needed here? or do we want to return resolved addresses like the snmp version does?

lib/App/Netdisco/Transport/CLI.pm
```diff
@@ -0,0 +1,84 @@
...
+    user => $selected_auth->{username},
+    password => $selected_auth->{password},
+    timeout => 30,
+    async => 0,
+    default_stderr_file => '/dev/null',
+    master_opts => \@master_opts
+  );
+
+  my $CONFIG = config();
+  $Net::OpenSSH::debug = ~0 if $CONFIG->{log} eq 'debug';
```
inphobia: do we want $ssh or the perl class here?

inphobia:
some side notes:
* `netdisco-do -e arpnip -DI` has a lot of different output for a cli collected device vs an snmpv3 one (cli is a lot more verbose). perhaps netdisco-do should have an -O option like you added in #460 ? or perhaps net::openssh debug should only kick in from -DD ?
* check_mac doesn't report device ip: `[13541] 2019-02-23 04:15:25 debug  [] check_mac - HSRP mac [00:00:0c:9f:f0:fe] - skipping`
* snmp arpnip seems to run after cli arpnip. could be because i have 2 matching device_auth entries though. did remove cronjob & old sshcollector config:
```
[13541] 2019-02-23 04:15:25 debug  [] check_mac - HSRP mac [00:00:0c:9f:f1:68] - skipping
[13541] 2019-02-23 04:15:25 debug  [] check_mac - HSRP mac [00:00:0c:9f:f1:7c] - skipping
[13541] 2019-02-23 04:15:25  info  [x.x.x.x] arpnip cli - Ended arpnip cli for x.x.x.x
# DESTROY(Net::OpenSSH=HASH(0x56135bd06558), pid: 13546)
# sending exit control to master
# call args: ['ssh','-O','exit','-T','-S','/home/x/.libnet-openssh-perl/b932996819e837026e81358b09021f45','-l','x','x.x.x.x','--']
# open_ex: ['ssh','-O','exit','-T','-S','/home/x/.libnet-openssh-perl/b932996819e837026e81358b09021f45','-l','x','x.x.x.x','--']
# io3 mloop, cin: 0, cout: 1, cerr: 0
# io3 fast, cin: 0, cout: 1, cerr: 0
# stdout, bytes read: 20 at offset 0
#> 45 78 69 74 20 72 65 71 75 65 73 74 20 73 65 6e 74 2e 0d 0a                                     | Exit request sent...
# io3 fast, cin: 0, cout: 1, cerr: 0
# stdout, bytes read: 0 at offset 20
# leaving _io3()
# _waitpid(13549) => pid: 13549, rc: 0, err:
# set_error(1 - aborted)
# master state jumping from _STATE_RUNNING to _STATE_KILLING
# killing master
# _master_kill: 13546
# waitpid(master: 13546) => pid: 13546, rc:
# master state jumping from _STATE_KILLING to _STATE_GONE
[13541] 2019-02-23 04:15:25 debug -> run worker main/nodes/100
[13541] 2019-02-23 04:15:25 debug skip: namespace passed at higher priority
[13541] 2019-02-23 04:15:25 debug -> run worker main/subnets/100
[13541] 2019-02-23 04:15:25 debug snmp reader cache warm: [x.x.x.x]
[13541] 2019-02-23 04:15:25 debug [x.x.x.x:161] try_connect with ver: 3, class: SNMP::Info::Layer3::Nexus, comm: <hidden>
SNMP::Info::_global uptime : DISMAN-EVENT-MIB::sysUpTimeInstance : .1.3.6.1.2.1.1.3.0
SNMP::Info::_global layers : SNMPv2-MIB::sysServices.0 : .1.3.6.1.2.1.1.7.0
SNMP::Info::_global description : SNMPv2-MIB::sysDescr.0 : .1.3.6.1.2.1.1.1.0
SNMP::Info::_global id : SNMPv2-MIB::sysObjectID.0 : .1.3.6.1.2.1.1.2.0
SNMP::Info 3.64
SNMP::Info::device_type() layers:01000110 id:9 sysDescr:"Cisco NX-OS(tm) nxos.7.0.3.I7.5a.bin, Software (nxos), Version 7.0(3)I7(5a), RELEASE SOFTWARE Copyright (c) 2002-2018 by Cisco Systems, Inc. Compiled 10/12/2018 18:00:00"
SNMP::Info::_load_attr old_ip_table : IP-MIB::ipAdEntAddr : .1.3.6.1.2.1.4.20.1.1
SNMP::Info::_load_attr old_ip_netmask : IP-MIB::ipAdEntNetMask : .1.3.6.1.2.1.4.20.1.3
SNMP::Info::_load_attr udpLocalAddress : UDP-MIB::udpLocalAddress : .1.3.6.1.2.1.7.5.1.1
[13541] 2019-02-23 04:15:26 debug  [x.x.x.x] arpnip - found subnet x.x.x.x/24
```
(some entries obfuscated, but meh)
* lib::App::Netdisco::Transport::CLI.pm should be mentioned in Build.pl/meta.yml/wherever the build scripts expect it.


that & the few review notes seem to be all the possible issues i see however.

ollyg:
I've created a branch and merged the PR into that, to continue working more easily in the main repo.

Agree with @inphobia comment above about "want to return resolved addresses like the snmp version does?" so I will take a look and do that. Also tidy and clean the code a bit.

Leaving PR open for now.

rc9000:
Ok cool, thanks! I'll add further changes in the branch here. BTW I also wrote the deprecation page that is mentioned in the sshcollector binary: https://github.com/netdisco/netdisco/wiki/netdisco-sshcollector-Deprecation

lib/App/Netdisco/Transport/CLI.pm
rc9000: Not sure if it makes a difference, this is straight out of the Net::OpenSSH documentation.

inphobia: ah yes, i see now how it works. perhaps this is better at home at line 59, so you get the output of the ssh master connection as well.

Issue #491, opened 2019-01-19
portcontrol uses vlan_reconfig_check instead of port_reconfig_check


inphobia:
https://github.com/netdisco/netdisco/blob/master/lib/App/Netdisco/Worker/Plugin/PortControl.pm#L20
https://github.com/netdisco/netdisco/blob/master/lib/App/Netdisco/Worker/Plugin/Power.pm#L22

i think vlan_reconfig_check might be a typo, don't we need port_reconfig_check instead?

it seems this makes port up/down & power control from the webinterface not work.

## Current Behavior

vlanctl config setting influences port power control & port bouncing behaviour.

## Possible Solution

use port_reconfig_check instead of vlan_reconfig_check?

## Steps to Reproduce (for bugs)

1. set vlanctl: false in config
2. try to change poe status of a port

## Your Environment

* Netdisco version used: 2.40.3
* SNMP::Info version used:

PR #492, opened 2019-01-19
fix for #491 (untested)

inphobia:
should be working according to stevenr on irc.

also move 1 sprintf from single to double quotes, else we don't print device i think.


inphobia:
should flatten my forked commit log, most of the commit messages are besides the point.

PR #493, opened 2019-01-25
i think end anchor should be in the if/end clause

inphobia:
if i'm reading the code correctly that is....


ollyg:
well spotted, thanks

Issue #510, opened 2019-02-17
extracting device_port ifindex() from ND database? Does it exist?


Stoatwblr:
I'm kludging together a script as first pass at demonstrating use of ND as source for GLPI/fusioninventory network inventory, as the existing fusioninventory code is "somewhat broken" (it only really works for Catalyst switches and having multiple sets of software doing polls isn't a good idea anyway as a lot of switches are quite fragile)

This uses OCSinventory XML format - The format is at http://fusioninventory.org/documentation/dev/spec/protocol/netinventory.html (but not well explained)

Unfortunately... (there's always a but isn't there?) I can't find any column in device_port_* or anywhere else which is related to ifindex. 

A lot of third party stuff is pinned on using ifindex numbers so keeping this intact would be preferable to reconstructing it later if feeding to third party software.

rc9000:
Unfortunately currently the ifindex is not stored in the database. You can however display them live from the device with the show command:

     $ bin/netdisco-do show -d 10.5.43.29 -e interfaces 
    [11825] 2019-02-18 10:06:31  info App::Netdisco version 2.040002 loaded.
    [11825] 2019-02-18 10:06:31  info show: [10.5.43.29]/interfaces started at Mon Feb 18 11:06:31 2019
    \ {
        1    "GigabitEthernet0/0",
        2    "Null0",
        3    "GigabitEthernet1/0/1",
        4    "GigabitEthernet1/0/2",
       ....

Maybe you can use this with some scripting as a workaround? Long-term I'd say that this is a worthwhile feature request, we've also been missing the ifindex already when linking Netdisco data to other systems.

Stoatwblr:
I've been discussing this with Oliver and JVI in the IRC channel today. They're agreeable to the idea.

The idea is to avoid adding extra load to the switches. For the moment I'm extracting the ifindex (and in/out octets/errors) from mrtg configfiles and rrd databases - polling them for the ifindex is only a few seconds but the counters can be as much as 4 minutes per pass (Huawei Quidways - they have awful SNMP code and an attitude of "nobody really uses SNMP, do they?", so they farm it out to "low cost Mumbai contractors"....) and I'd prefer to avoid abusing the switches too much as the cpu load goes through the roof.

rc9000:
Ok I'll give it a try to implement that. No promises on an ETA though, might take me a while.

[...]

PR #513, opened 2019-02-21
PR for #510 - store ifindex in device_port

rc9000:
* adds column device_port.ifindex (bigint)
* extend Discover/Properties.pm to store it
* note that this is only a DB change, currently no support in the UI


share/schema_versions/App-Netdisco-DB-53-54-PostgreSQL.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE device_port ADD COLUMN "ifindex" bigint;
```
ollyg: I think this is better off in the device_port_properties table, which is storing the 'raw' data for other fields too.
```suggestion
ALTER TABLE device_port_properties ADD COLUMN "ifindex" bigint;
```

lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
```diff
@@ -229,6 +229,7 @@ register_worker({ phase => 'early', driver => 'snmp' }, sub {
           type         => $i_type->{$entry},
           vlan         => $i_vlan->{$entry},
           pvid         => $i_vlan->{$entry},
+          ifindex      => $entry,
```
ollyg: This will also need changing to use device_port_properties

lib/App/Netdisco/DB/Result/DevicePort.pm
```diff
@@ -69,6 +69,8 @@ __PACKAGE__->add_columns(
   { data_type => "integer", is_nullable => 1 },
   "lastchange",
   { data_type => "bigint", is_nullable => 1 },
+  "ifindex",
```
ollyg: this will also need changing to use Result/DevicePortProperties.pm

share/schema_versions/App-Netdisco-DB-53-54-PostgreSQL.sql
rc9000: Alright, I wasn't quite aware what should go into _port and _port_properties respectively. I'll update the PR

rc9000:
Ok it's now moved to device_port_properties. Since this seems to be a pretty harmless change, I'll just go ahead an merge it. Let me know if there's something else I should do otherwise, I'm very new to developing Netdisco contributions still :)

Issue #514, opened 2019-02-25
reports->ip->ip inventory inconsistent results


inphobia:
reports -> ip -> ip inventory is missing mac addresses for some ip's at random intervals. seems very similar to #274 

## Expected Behavior

show mac address if it's somehow known.

## Current Behavior

well, this one is curious, if you just press "search ips" or refresh the browser page with ctrl+f5 sometimes the mac addresses do show up.
using default settings: 1970-01-01 -> today timerange

i searched for 10.98.17.0/24 3 times, only paused to take the screenshots. i sorted on mac address so the missing addresses would show up on top.

in a span of 10 seconds it first returned 4 blank mac entries, then 2, then 3.

4 macs missing:
![image](https://user-images.githubusercontent.com/39005454/53308183-5f386e00-389f-11e9-9aae-070a47c1ee37.png)

2 macs missing:
![image](https://user-images.githubusercontent.com/39005454/53308185-6495b880-389f-11e9-837a-addb2fe9b031.png)

3 macs missing:
![image](https://user-images.githubusercontent.com/39005454/53308187-68293f80-389f-11e9-8a55-480780659aa8.png)

i did notice that ip's with no macs almost always are devices known in the database. for nodes it always seems to return the mac if it is known.

* it is not limited to 1 devicetype or operating system version. we see this on ios router, ios switches, cisco small business switches, aerohive access points, ...
* operating system versions are divers as well.
* we're also running snmp arpnip & netdisco-sshcollector (testserver is running the version @rc9000 is working on, production server uses the default one)

strangly enough the ip<->mac mapping is in node_ip, so i suspect this report is also sourcing data from other tables. i checked node_ip for 2 of the devices you saw in the screenshots above (10.98.17.200 & 201), they only have 1 entry for their ip & mac, and it's labeled active (so no archived data present)

*** side note: the expire job will completely remove all device entries in node_ip. when the next arpnip runs on the subnet they will be added again.

the mac address stored in node_ip is not present in node, but that makes sense since devices don't have a node entry. since no entry is present in node the expire job will delete 

i also verified that nothing was updating the database at the time i tested:
 * ran discoverall, macwalk & arpwalk - waited for them to complete
 * deleted jobs via webinterface to make sure nothing was stuck
 * shut down & restarted netdisco
 * waited for devices with snmp failures to get deferred again (since after a restart the get another try)

## Possible Solution

as noted in #274 , limiting the timeframe seems to lessen the variations.

[...]

PR #517, opened 2019-02-27
always prefer device entries over node_ip entries

inphobia:
first part of the fix for #514 

select distinct needs to be provided with an explicit order to return consistent results. ip & active suffices for nodes, but if it's a device with an entry in node_ip we get random results. so also order on boolean node to prefer device entries in that case.

(at the time of writing however none of my devices return a mac address, which might be another bug)


ollyg:
Closing PR as Github seems to be broken and not allow me to activate the draft PR.

Have applied change in:
[master 77decc23] #514 inconsistent results in ip inventory (inphobia)

PR #532, opened 2019-03-11
Support for external command to provide device_auth stanza

ollyg:


rc9000:
That's excellent, thanks!

PR #538, opened 2019-03-13
Add store phase and update Arpnip::Nodes to use it

ollyg:


rc9000:
Awesome @ollyg ! 

I didn't mean that you had to implement that just for my request, nor on the same day on top of that, but hey I'll take it. Thanks!

PR #539, opened 2019-03-14
Fix an issue with the reindixing for VRF

earendilfr:
When we perform the reindexing for the VRF, we need to avoid to use the prefix vlan- for SNMP context.

### DIFF
--- a/lib/App/Netdisco/Util/SNMP.pm
+++ b/lib/App/Netdisco/Util/SNMP.pm
@@ -100,11 +100,19 @@ sub snmp_comm_reindex {
       }
       $prefix ||= 'vlan-';
 
-      debug
-        sprintf '[%s] reindexing to "%s%s" (ver: %s, class: %s)',
+      if ($vlan =~ /^[0-9]+$/i && $vlan) { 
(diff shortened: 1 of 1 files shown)


ollyg:
Hi @earendilfr thanks for the patch! As I do not have VRFs to test on, I would like to understand how this is meant to be used (also so I can update the docs).

It seems that the patch supports three modes:
* numeric vlan (old style working
* alphanumeric "vlan" (this is the one you added? can you give an example?)
* no vlan or zero (to reset)

As above can you give examples of how this is used by you?

earendilfr:
Hello,

The use is case is for VRF (at least for Cisco devices).
When you want to perform a ARP discover with multiple VRF on a devices, you need to :
- create on the cisco devices a new SNMP context for this vrf with the same name than the VRF:
`snmp-server context vpc-keepalive vrf vpc-keepalive`
- after the that, the discover of the IP addresses and the ARP entries for VLAN inside this VRF will be discovered automatically.

ollyg:
thanks! I have updated the Vendor Tips documentation.

PR #548, opened 2019-03-23
Add IS-IS Support

### LINKED ISSUE #546: isis neighbor discovery?
do we need is-is handling in neighbor discovery?

https://github.com/netdisco/netdisco/blob/master/lib/App/Netdisco/Worker/Plugin/Discover/Neighbors/Routed.pm

netdisco/snmp-info#316 - @pyro3d 

## Your Environment

* Netdisco version used: HEAD
* SNMP::Info version used: HEAD

pyro3d:
Add IS-IS support. Needs SNMP::Info to also support IS-IS. 
fixes #546

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -75,7 +75,7 @@ Module::Build->new(
     'Starman' => '0.4008',
     'Storable' => '0',
     'Sys::SigAction' => '0',
-    'SNMP::Info' => '3.65',
+    'SNMP::Info' => '3.66',
     'SQL::Abstract' => '1.85',
     'SQL::Translator' => '0.11024',
     'Template' => '2.24',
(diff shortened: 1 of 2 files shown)


ollyg:
Thanks! Question: why have both `isis_peers` and `isis_routers` but we only use `isis_peers`?

inphobia:
> Question: why have both `isis_peers` and `isis_routers` but we only use `isis_peers`?

while i can't speak for @pyro3d , the code seems to be modeled after the ospf discovery where both local and remote ids seem to be used to queue for discovery.

after having a look at rfc4444 it seems that isis_adj_id() is most likely not usable, since that gets isisISAdjNeighSysID values which is a 6byte string.

pyro3d:
Yeah, I modeled it off OSPF. Originally I was going to use the ISIS adj id as well for discovery, however it's not formatted properly (6 byte string vs 4 byte string), and doesn't always contain an IP address. I've removed it from the patch.
Example:
IPv4 Loopback
  10.13.37.1
ISIS CLNS address (per standard formatting, see https://www.juniper.net/documentation/en_US/junos/topics/concept/is-is-routing-overview.html#id-11020505 )
  49.0001.**0100.1303.7001**.00

```
ISIS-MIB::isisISAdjNeighSysID.522.1 = Hex-STRING: 01 00 13 03 70 01

ISIS-MIB::isisISAdjIPAddrAddress.522.1.1 = Hex-STRING: 0A 14 1E 01
ISIS-MIB::isisISAdjIPAddrAddress.522.1.2 = Hex-STRING: FE 80 00 00 00 00 00 00 02 12 1E FF FE 23 94 DE
```

PR #564, opened 2019-04-23
Modification of netdisco-do 

earendilfr:
All to indicate the "-d" parameters multiple time to perform an action on many host with only one command call.


ollyg:
@earendilfr many thanks!!

inphobia:
does this need special handling for ´renumber´?

earendilfr:
I didn't have check for renumber but I think it could be a good idea to add an option to avoid multiple -d option in this case...
I can perform the modification tomorrow.

PR #570, opened 2019-04-30
while we target perl 5.14, we can test a few more versions too

inphobia:


inphobia:
add tests for perl 5.20 & 5.26 too, first run completed succesfully

Issue #594, opened 2019-05-26
device_port_log table does not have a primary key


inphobia:
but according to the docs it should have one. fix the docs or change the database?

https://github.com/netdisco/netdisco/blob/065e74df62db74064776b0b40fd65f31a035e1d1/lib/App/Netdisco/DB/Result/DevicePortLog.pm#L43

ollyg:
thanks, applied fix in
[master 744da684] #594 add PK to device_port_log and other tidying

PR #595, opened 2019-05-26
dbic sync between actual db & our functions/docs

inphobia:
still haven't figured out how to magically make all those database .pm files we have (table definitions i've got working, but functions & all other stuff is still a mystery). that did turn up some inconsistencies between the actual db schema & the db modules we have.
i've taken the schema as information source & modified our modules to they match.

for device_port_ssid & _wireless i also added a primary key to the db. it was in the docs but not in the schema.

also deleted DeviceRoute.pm, a module for a non existing table.

important: could use some more testing!


inphobia:
still to fix: #594 
and also related  #596

ollyg:
looks good, thanks!

PR #602, opened 2019-06-02
html prefixup

inphobia:
we're moving to html5 .... someday

clean up some cruft for both xhtml1 & html5 to get sources cleaned up

also: qw() -> qw// changes

some of these changes are mostly cosmetic, but since i'll be relying on intellij for quite some linting of html, css & js, i'm pushing these changes too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ _build
 blib
 !.docker/hooks/*
 .idea
+/node_modules/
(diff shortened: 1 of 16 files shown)


inphobia:
https://github.com/inphobia/netdisco/tree/nn_preformat will replace this one after i do some testing

Issue #610, opened 2019-06-25
Managed to crash netdisco weboutput


zoeloe:
I'm not proud of it, but I think a certain search query crashes netdisco weboutput.

## Expected Behavior
Resist crashing when using malformed searches

## Current Behavior
Crash, generated weboutput wrong
crashdump in gist:
https://gist.github.com/zoeloe/086bf4b10ceace6f5f1c6c389a45195e

## Possible Solution
Filter malformed queries?

## Steps to Reproduce (for bugs)
1. http://<host.domain.part1.td>/search?q=2C101%2FA%2F5+2C101+++2C101%2FA%2F6+2C101&firstsearch=on

## Context
pasted a copy buffer by accident in the main search field on home page. 

## Your Environment
Ubuntu 16.04.05
* Netdisco version used: 2.42.10
* SNMP::Info version used: 3.68

## Device information
not relevant.

inphobia:
it's indeed easy to reproduce. a fix however seems less evident since the search bar tries to play it smart to figure out if you're looking for vlans, nodes, devices or ports. so fixing the mac detection seems to have unintended side effects.

general input sanitation seems to be a good path to take, that or better error handling. would love some input from @netdisco/netdisco-admins / @netdisco/netdisco-developers so see what their take is.

(side note, am slowly working on a frontend rewrite, this could be a target for that too)

inphobia:
trouwens, blij om te zien dat de vub bij onze gebruikers zit :)

zoeloe:
@inphobia 
Nice to know that there is a dutch speaking (Like me !) and a major dev/contributor on the netdisco project.
The only mini contrib from me was in #239 for SNMP::Info.
And some bug reports which I bumped into.

You can leave your personal coordinates for correspondence in the dutch language if you like, for me.
(replied in English for general readability :) )

Yes, I work at the VUB university and promoted the use of netdisco from the early existence on, because of the (IMHO) layered (SNMP::Info abstraction layer) philosophy as starters.

Keep up the great work. I certainly appreciate.

Regards, Thanks for your contributions.

JeroenvIS:
@zoeloe inphobia isn't the only Dutch speaking dev on the project ;-) (though I'm not contributing much these days...)
But indeed, let's try to keep the tickets and other general correspondence in English.

JeroenvIS:
[...]

PR #611, opened 2019-06-26
compare ieee mac against 6 iterations

inphobia:
perhaps there is a reason to use '00:00:00:00' instead of '00:00:00:00:00:00', but i can't see why, even when going through git history.


inphobia:
perhaps the man, the legend @ollyg has a clue on why this could be?

(found when reviewing #610)

PR #621, opened 2019-07-16
Modified to add Radius support

jngreene:
Lines 110-112 commented out
Lines 113-124 added
Lines 233-249 added match_with_radius  function

### DIFF
--- a/lib/App/Netdisco/Web/Auth/Provider/DBIC.pm
+++ b/lib/App/Netdisco/Web/Auth/Provider/DBIC.pm
@@ -15,6 +15,10 @@ use Digest::MD5;
 use Net::LDAP;
 use Try::Tiny;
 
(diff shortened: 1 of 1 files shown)


ollyg:
Thanks @jngreene this is a great idea! I'd like to tidy up the code a little before merging, which I will take care of for you.

inphobia:
had started this myself (the cleanup), but really start to set priorities instead of starting everything but never finishing...

2 takeaways:
* dont forget to add `Authen::Radius` to Build.PL
* this is also an annoying module imo, since it expects to have its radius dictionaries in /etc, so i think we need some way to keep these somehow contained in the netdisco distribution like we do for all other files...

i briefly considered trying to make it into an external module like rancid & observium, but i think we're past the point of trying to keep bloat & dependencies in check :)

PR #629, opened 2019-07-31
Allow to use a special identity file to connect to device in SSH

earendilfr:
For the SSH collector, we cannot specify a specific identity file for SSH.
With this, we can indicate an identity file and a passphrase for the.

Bellow the key to add to a "device_auth" block:
- identity_file: to indicate an identity file
- identity_pass: to indicate the passphrase for this file

Just a question: i din't havce found if the mac page for App::Netdisco::Manual::Configuration is generated. Did I need to update it ? If yes, where is located the file that generate the man page?

### DIFF
--- a/lib/App/Netdisco/Transport/SSH.pm
+++ b/lib/App/Netdisco/Transport/SSH.pm
@@ -79,6 +79,8 @@ sub session_for {
     $device->ip,
     user => $auth->{username},
     password => $auth->{password},
+    key_path => $auth->{identity_file},
+    passphrase => $auth->{identity_pass},
     timeout => 30,
     async => 0,
     default_stderr_file => '/dev/null',


rc9000:
Hi @earendilfr 

Regarding the manual, this is now only in the Wiki. If your system still has this manpage, its from an earlier release (these POD files got removed in 8b1b97f58c9f).

Regarding the patch, it is already possible to supply the identity file with -o IdentityFile=xxx in device_auth.ssh_master_opts. But I'm not against including your easier method and document this one instead - it might not be very intuitive how it is now, as proven by you making this PR :)

For the passphrase, I think the industry best practice would be to use ssh-agent(1) instead of storing passphrases in clear text, but I'm also on board with including it if it covers and actual need for people. It's maybe still a bit ahead of using passwordless keys.

I'll leave the final merge decision to @ollyg & co.

Cheers
Christian

ollyg:
Agree with @rc9000 that storing the passphase in Netdisco's config isn't a good idea. I think booting up ssh-agent externally to Netdisco and using that would be better.

inphobia:
as much as i agree with not storing passwords or keys plaintext, the same argument could be made for all other credentials which we save in deployment.yml. you know you can also create a read only ssh user. plain text ssh passwords in the config file are still more secure as snmp v2 or v3 without crypto.

(deployment also referenced the nonexisting man pages btw, was fixed in 5fb9d52278c66402c69060d09cfc71ea4d105b1e)

my take on it is, since net::openssh which we use mostly lets the openssh binary do most of the work, we might as well use the .ssh/config for everything we can, like usernames, identity files, key specs for all crappy devices, etc...

counter argument is that this is fragmenting the config, and i do like the single config file we have.

but i've noticed someone already has a master plan for this 💡 https://github.com/netdisco/netdisco/compare/og-get_external_credentials 

as much as i like ssh for so many reasons, doing ssh safely is harder as it seems. pubkey scales just as badly as most other auth methods imo. you still need to have a way to disable compromised keys, you need to keep track which keys are allowed where, and how about periodic rotation (or do you still use that 512bit key from 10 years ago 🙈 ) ofcourse there are solutions for all of this (like https://github.com/AndriiGrytsenko/openssh-ldap-publickey ) but as strange as i find myself saying this, i'd actually for proposed patch.

or we can just use debian & don't worry about it :)
https://www.debian.org/security/2008/dsa-1571

PR #633, opened 2019-08-16
Change Port Free icon to be more accessible

hllnS:
The port free icon currently only differs from the port down icon by colour, on higher resolutions with a colour vision impairment they're not very distinguishable, a slightly different icon fixes this.


ollyg:
Thank you David, very much appreciated!
I'll try to squeeze out a release soon with this.

Issue #18, opened 2014-08-27
Add SNMP trap-receiver


netdisco-automation:
A port that comes up might trigger a discovery for other devices attached to that port (maybe with a certain delay?).
In this context it might also make sense to have some message-pane / log-window that will show which port came up/went down, with crosslinks to the device/port etc.

Reported by: neufeind

Original Ticket: [netdisco/nd2-features/18](https://sourceforge.net/p/netdisco/nd2-features/18)

ollyg:
Moved to wishlist page in wiki

olex777:
Are this case about snmp-traps was solved any of the ways? Where could I see it? Will be very thankfull for replay.

JeroenvIS:
This case isn't about a bug, it's a suggestion to extend current functionality. Therefore the suggestion has been added to the "wishlist" page on the wiki. As far as I know, there is nobody working on this wish.

djzort:
Perhaps create a script for SNMPTT -> http://snmptt.sourceforge.net/docs/snmptt.shtml#Executing-an-external-program

olex777:
Thank you, I think it will be usefull!

PR #634, opened 2019-08-20
Fix nbtstat jobs getting scheduled for devices

rc9000:
Hi @ollyg & team

I noticed that despite variations of `nbtstat_only: !any` and `nbtstat_no: any`, I kept getting a lot of

    nd2: #18 poll: #175319: nbtstat 191.168.7.91 

lines in my ps output, with the IP address being a known device. Looking at Worker/Plugin/Nbtstat.pm it seems to use is_macsuckable instead of is_nbtstatable. 

The fix seems to work fine here, but maybe there's some intent here that I don't understand, so I chose to make it a pull request instead of just committing. 

Cheers
Christian


ollyg:
Hi, so it's a little more complex than it might seem... 

What's happening is that NetBIOS discovery happens against nodes, not devices. So the nbtstat ACLs are IP ranges of nodes, which is why they don't affect the observed job, which is for a device.

The macsuck ACL is used because if we don't macsuck a device, then there are no nodes to check. And the nbtstat job is run against a device, to get the list of nodes to check.

So the job might be working just fine. Do you want nbtstat to run, or not? What is it you want to happen, then we might look at changing the way the options/config works, to be better.

rc9000:
Oh ok I see.  I wanted fully disable nbtstat, we don't want to interfere with the connected nodes - not our business.

But if I'm getting this right, you're saying the nbtstat <device> will get the list of all nodes, then my nbtstat_no: any will be honored and none of the nodes will be touched. That's exactly what I need, I just didn't understand why I'm getting all these jobs scheduled anyways.

Is there an easy way to also opt out of the nbtstat <device> action? It just clogs up the job queue in our case. But no worries if that's not the case, they seem to finish quick.

I'll remove the PR, thanks for the explanation!

ollyg:
I think this will work. It will be merged with the default config.

````
schedule:
  nbtwalk: undef
````

**Comment by ollyg** [comment]
(note it should be null, not undef!)

**Comment by inphobia** [comment]
this will indeed work, we've been running this for at least a year i think:

```yaml
schedule:
  discoverall:
    when: '5 3,7,11,15,19,23 * * *'
  macwalk:
    when:
      min: 55
      hour: '*/4'
  arpwalk:
    when:
      min: 5
      hour: '*/4'
  nbtwalk: null
  expire:
    when: '30 23 * * *'
  makerancidconf: null
```

PR #637, opened 2019-08-21
fix raidus typo

alanbuxey:
radius not raidus


ollyg:
Thanks!! Really pleased to see people looking at the code already. Shame I didn't test it as it was done on a moving train :-D

PR #640, opened 2019-08-29
some documentation fixes

inphobia:
* the pgtune we suggested seemed to be quite dead, provide a few working alternatives
* make mention of release notes in netdisco-deploy, since incompatible changes are listed there
* steal pod test from snmp::info, to root out most basic pod errors
* netdisco-deploy had a shadowed variable, rename the second copy (found by pycharm/intellij idea)
* regen dist files while here


inphobia:
dunno what happened with commit 4cc09bc but since then all my merges from master get logged. need to figure out how to fix that without destroying my branches.

PR #655, opened 2019-09-29
jquery security fix for CVE-2019-11358

inphobia:
https://nvd.nist.gov/vuln/detail/CVE-2019-11358#vulnCurrentDescriptionTitle
since we are some versions (jq 1.9.1 or something close to that according to the source) behind the last release of the 1.x train (1.12.4) the patch was not compatible, so the file has been manually edited to apply the needed changes

jquery 1.12.4 official patch:
https://github.com/DanielRuf/snyk-js-jquery-174006/blob/master/jquery-1.12.4.min.patch


inphobia:
the major reason why the upstream patch isn't compatible, even with fuzzy patching, is because somewhere along the lines all the variable names also have changed.

i've been working on and off to move us to move us to frameworks for css & js, so that css compiling, js minimizing, whatever will be used instead, mostly for easier tracking of  local changes, keeping libraries currect & as a way to be able to support nd2 with css/js (perhaps a few html things too) if we ever go into maintenance mode when nd3 becomes a thing.

note: the goal is to still have compiled/minimized files only in the distribution, so users won't need to worry about any of that.

PR #680, opened 2019-11-07
Reduce macsuck bandwidth usage to database

rc9000:
From IRC today:

    19:31 < rc9000> hi guys! some days ago our db cluster switched my postgres instance 
                    to the backup dc, but netdisco remained at the main dc.
    19:31 < rc9000> a little later I got complaints that I was using about 400mbit/s of 
                    the already highly saturated link
    19:32 < rc9000> it turns out that get_port_macs queries a full list of all device_port 
                    ips and macs, which is about 13MB in our case
    19:33 < rc9000> and we poll 9000ish devices in 40 minutes, so the math works out :) 
    19:34 < rc9000> i'm looking at the code and it seems i could make a slightly more specific 
                    get_port_macs inside walk_fwtable, that contains only the macs really 
                    returned from snmp->fw_mac

This PR tries to fix the above with a get_port_macs that's called inside walk_fwtable and only returns the macs found in the current target device/vlan.

If I'm reading walk_fwtable right the only lookups that can ever be made are for macs that the current snmp->fw_mac contains, but I'd appreciate if @ollyg and other cracks could verify this assumption. Also, I did not find usages of get_port_macs outside of Macsuck/Nodes.pm, so I hope this doesn't break anything.

I'm trying it now in our environment and it both seems to work ok so far and has reduced bandwidth DB -> poller by orders of magnitude.

### DIFF
--- a/lib/App/Netdisco/Util/PortMAC.pm
+++ b/lib/App/Netdisco/Util/PortMAC.pm
@@ -5,7 +5,7 @@ use Dancer::Plugin::DBIC 'schema';
 
 use base 'Exporter';
 our @EXPORT = ();
-our @EXPORT_OK = qw/ get_port_macs  /;
+our @EXPORT_OK = qw/ get_port_macs /;
 our %EXPORT_TAGS = (all => \@EXPORT_OK);
 
 =head1 NAME
@@ -32,7 +32,6 @@ sub get_port_macs {
 
     my ($fw_mac_list) = @_;
     my $port_macs = {};
-    
 
     my $dp_macs
         = schema('netdisco')->resultset('DevicePort')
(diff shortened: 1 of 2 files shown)


inphobia:
at first glance seems to work just fine, but i'll need to wait a few days so i actually have users moving around etc...
(and also only have a few days of history atm, will see if i can copy production db to test & run with the same config for a few days)

rc9000:
Some related RRD:

Traffic before/after on database server switch port (hard to see now but there are some green pixels there)

rrdsgp48

Effect on poller performance

scr 2019-11-08 at 08 53 46

rc9000:
Performance implications of this PR from a single test over 7k devices, mostly Cisco:

 * 190502 queries performed for 7186 devices macsucked (26.5 vlans/queries per device)
 * max vlans in "in" clauses 868, this query takes 0.28 sec and finishes successfully.
 * average query takes 0.0304 sec.
 * 26 of the queries took >= 1 sec
 * all these queries together take 5806 sec

Full CSV this comes from:
[gm_profile_res.csv.zip](https://github.com/netdisco/netdisco/files/3830768/gm_profile_res.csv.zip)

PR #695, opened 2020-01-23
new SSHCollector platform for Extreme VSP

antonc42:
This module works for Extreme (formerly Avaya) VSP switches running the
VOSS operating system. When using multiple VRFs, the default Arpnip does
not work.

The way to query the ARP tables of different VRFs is to append "::X" to
the end of the SNMP community string, where "X" is the VRF ID number.
Since Netdisco doesn't support querying with multiple community strings
at this time, this module uses SSH to query the ARP tables of all the
VRFs by default or select VRFs using the "vrfs" key in "device_auth"
(see documentation in the module for an example).

### DIFF
--- a/lib/App/Netdisco/SSHCollector/Platform/VOSS.pm
+++ b/lib/App/Netdisco/SSHCollector/Platform/VOSS.pm
@@ -1,0 +1,134 @@
+package App::Netdisco::SSHCollector::Platform::VOSS;
+
+=head1 NAME
+
+App::Netdisco::SSHCollector::Platform::VOSS
+
+=head1 DESCRIPTION
+
+Collect ARP entries from Extreme VSP devices running the VOSS operating system.
+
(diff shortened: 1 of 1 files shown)


antonc42:
We have it running against several different VSPs right now and it seems to be working fine. We have VSP 8400, VSP 7200, and VSP 7400 models running VOSS 7.1 and 8.0.

mgpl2001:
Hello, can the ssh collector for Cisco IOS also do something similar? The network I work on has lots of Cisco IOS routers with VRFs. We don't necessarily need full VRF support in Netdisco but currently cannot find devices like we do on networks without VRFs.  It is hit or miss.  I am not sure if I understand how this is working since plain SNMP collectors have been working well for us for many years (except for the VRFs). I would be very interested if there is an existing solution. I will try to read the code for the Extremes to see if I understand what you are doing.  Thank you for Netdisco, it is a wonderful application.

PR #714, opened 2020-04-19
Add vendor to IP Inventory report

rc9000:
* request on mailing list by Adrian Sevcenco, 08.04.20,
   "[Netdisco] how to modify report: add column to ip inventory"
 * I think there is no harm in adding this generally to the report, but let me know if you think otherwise


ollyg:
looks good to me! 👍

PR #725, opened 2020-06-04
Avoid duplicate device in rancid export file

earendilfr:
- Add a variable to check that a device is unique in rancid exported file
- Add a default domain_suffix to avoid some warning message


ollyg:
thanks @earendilfr! I will release it now to CPAN

Issue #774, opened 2021-03-07
Crypt::CBC 3.01 fills up backend logs


inphobia:
https://metacpan.org/pod/Crypt::CBC 3.01 has been released which wants to get rid of "opensslv1" as key derivation function, basicly md5 based.

```
        - Warn when the deprecated opensslv1 PBKDF (key derivation function) is used
          for encryption. Turn off with -nodeprecate=>1 or by choosing a different
          PBKDF, such as -pbkdf=>'pbkdf2'.
```

now i'm not sure if it's Session::Storage::Secure or Dancer::Session::Cookie where the problem lies, but my logs are filling up with:

```
WARNING: The key derivation method "opensslv1" is deprecated. Using -pbkdf=>'pbkdf2' would be better.
Pass -nodeprecate=>1 to inhibit this message.
 at /home/netdisco/perl5/lib/perl5/Session/Storage/Secure.pm line 213.
```

due to the layers of abstraction i don't see an easy way to change either of the possible offenders quickly:
https://github.com/dagolden/Session-Storage-Secure/blob/8b16fba0323dbf0d41a90cc290932d675a30f63c/lib/Session/Storage/Secure.pm#L210 (and a few other places)
https://github.com/PerlDancer/Dancer-Session-Cookie/blob/dd912f05a0ac8345a412c5aa4cbd936f95ef523e/lib/Dancer/Session/Cookie.pm#L43

could be as easy as passing a `-pbkdf` to `Crypt::CBC->new`, or ignoring the warning with `-nodeprecate`.

(crypt::cbc says it's 100% perl so we don't need to worry about openssl versions)

inphobia:
this was fixed by the recent release of session::storage::secure 1.000
https://github.com/dagolden/Session-Storage-Secure

https://github.com/dagolden/Session-Storage-Secure/issues/8

netdisco pulls in s::s::s via dancer::session::cookie:
https://github.com/netdisco/netdisco/blob/951139fe7c6511c344bd20b2543463a6828ff295/META.json#L55

adding an explicit req on session::storage::secure 1.000 should fix this, but i'm not clear on how this will affect older session cookies. s::s::s says it should be backwards compatible. 

with new netdisco deployments this works for me at least.

PR #781, opened 2021-04-13
require Session::Storage::Secure 1.000 for #774

inphobia:
needs more testing in upgrade scenarios.

also bump snmp::info to 3.71, 3.70 is 18months old now.


ollyg:
we can put in the release notes that sessions will be invalidated by the upgrade and users must log in again

inphobia:
> we can put in the release notes that sessions will be invalidated by the upgrade and users must log in again

made an entry in the release notes for `2.04700X` lest we forget.

https://github.com/netdisco/netdisco/wiki/Release-Notes#204700x

PR #782, opened 2021-04-14
Allow <"> in remote_port

rc9000:
Is it ok if we remove <"> from the "weird characters" in Neighbors.pm? Background is that the remote_port from Nokia TiMOS (the former Alcatel-Lucent brand devices) looks like this: 

    6/2/13, 10/100/Gig Ethernet SFP, "the descrption the user entered"

Not removing the <"> makes the device_port.remote_port match the matching device_port.descr, and also makes the "Connected devices" link in the Web UI work. I found no adverse effects.


rc9000:
Merging this since nobody seemed to have any concerns.

PR #794, opened 2021-06-17
replace unicode nonbreaking space with ascii space

inphobia:
all changes are in comments, so after a # , no actual code changes

for some reason unicode nonbreaking spaces have gotten into some code which i only noticed for the first time when trying an install on openbsd.

this changes them all into ascii spaces.

in vi you can see this by first setting a few options:
:set display+=uhex
:set isprint=
:set encoding=latin1

changing them can be done with sed:
sed -i $'s/\xC2\xA0/ /g' 

in a very select few files it can be that only u+00c2 or u+00a0 are present, these can be found with:
(backticks escaped with backslashes for github markdown formatting)

grep -rl -- \`printf '\xA0'\` .
grep -rl -- \`printf '\xC2'\` .

i didn't change imported javascript, css, etc....

### DIFF
--- a/Changes
+++ b/Changes
@@ -679,7 +679,7 @@
 
   [NEW FEATURES]
 
-  * #401 Autodiscovery via EIGRP peers
+  * #401 Autodiscovery via EIGRP peers
 
   [BUG FIXES]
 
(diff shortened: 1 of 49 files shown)


ollyg:
many thanks!

PR #815, opened 2021-09-12
pseudodevice updates

inphobia:
see irc for the reason why this is a draft request.

a few more updates to pseudo devices.
- give them some more info on creation (modelname, os, ports_nr, port names)
- set os version to the netdisco release they were created with
- export pretty_version  from utils/statistics to make that happen

there's more on my wishlist, but let's start with this.


ollyg:
I would also enable layer 2 on pseudo devices (layer 3 is currently the only option).

ollyg:
Thanks @inphobia and I've also fixed the skipping of contact|location|portname|snapshot for pseudo devices, in commit 6545fff0

PR #846, opened 2021-12-09
Ability to see all columns when searching device

ArthurHlt:
This PR add a `seeallcolumns` parameters in search device for seeing all columns from device.

Why is this useful ?
When using api, it miss some information like OS name. With this parameters all informations became available.

As I didn't wanted to break anything with webui, I've prefered to make this parameter.

I hope some others will find it useful.

BTW, thanks for this great tool


ollyg:
Thank you @ArthurHlt that's a great addition

PR #847, opened 2021-12-16
Do not macsuck suspended VLANs

dtuecks:
VLANs in the state 'suspended' cannot be polled via SNMP. While the VLAN state is suspended, the ports associated with this VLAN are shutdown, and that VLAN does not pass any traffic.
SNMP queries will result in a 'SNMP Auth Error' on the device and the polling takes longer (since Netdisco will retry the query a few times).
This pull request checks the VLAN state before macsucking and skips it if the state is 'suspended'.


ollyg:
thanks! have merged it

Issue #852, opened 2022-01-28
Error message when collecting ARP data from NXOS box using CLI


njohnsn:
## Expected Behavior

I'm trying to use the NXOS CLI to collect ARP and ND rather than SNMP

## Current Behavior

From the command line it appears to fail processing CLI entries and then defaults to SNMP.

Here is the output from the command line:

```
netdisco@netdisco:~$ ~/bin/netdisco-do arpnip -d 192.168.204.68 -DIS
[1482194] 2022-01-28 16:55:12  info App::Netdisco version 2.052000 loaded.
[1482194] 2022-01-28 16:55:12  info arpnip: [192.168.204.68] started at Fri Jan 28 16:55:12 2022
[1482194] 2022-01-28 16:55:12 debug => running workers for phase: check
[1482194] 2022-01-28 16:55:12 debug -> run worker check/_base_/0
[1482194] 2022-01-28 16:55:12 debug arpnip is able to run
[1482194] 2022-01-28 16:55:12 debug => running workers for phase: main
[1482194] 2022-01-28 16:55:12 debug -> run worker main/nodes/200
[1482194] 2022-01-28 16:55:12 debug cli session cache warm: [192.168.204.68]
[1482194] 2022-01-28 16:55:16 debug 192.168.204.68 1482194 arpnip()
[1482194] 2022-01-28 16:55:23 debug  [192.168.204.68] check_mac - VRRP mac [00:00:5e:00:01:14] - skipping
[1482194] 2022-01-28 16:55:23 debug  [192.168.204.68] check_mac - VRRP mac [00:00:5e:00:01:dc] - skipping
[1482194] 2022-01-28 16:55:23 debug  [192.168.204.68] check_mac - VRRP mac [00:00:5e:00:01:dd] - skipping
[1482194] 2022-01-28 16:55:23 debug  [192.168.204.68] check_mac - VRRP mac [00:00:5e:00:01:de] - skipping
[1482194] 2022-01-28 16:55:23 debug  [192.168.204.68] check_mac - VRRP mac [00:00:5e:00:01:df] - skipping
[1482194] 2022-01-28 16:55:23 debug  [192.168.204.68] check_mac - VRRP mac [00:00:5e:00:01:e0] - skipping
[1482194] 2022-01-28 16:55:23 debug  [192.168.204.68] check_mac - VRRP mac [00:00:5e:00:01:e1] - skipping
[...]

PR #854, opened 2022-02-01
Fix issue #852

njohnsn:
Fixes issue #852


rc9000 (approved):
Thanks @njohnsn, I was reading your issue and mailing list post and was about to try a similar fix. Much appreciated !

PR #937, opened 2022-10-21
Add macsuck worker to collect various PortAccessEntity (NAC) attributes

ollyg:


share/schema_versions/App-Netdisco-DB-75-76-PostgreSQL.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+ALTER TABLE device ADD COLUMN "pae_control" text;
```
ollyg, line 3: I think file `lib/App/Netdisco/DB/Result/Device.pm` needs to have the extra column added to spec?

lib/App/Netdisco/Worker/Plugin/Macsuck/PortAccessEntity.pm
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  # individual port properties
+  my $pae_authconfig_state = $snmp->pae_authconfig_state();
+  my $pae_authconfig_port_status = $snmp->pae_authconfig_port_status();
+  my $pae_authsess_user = $snmp->pae_authsess_user();
+  my $pae_authsess_mab = $snmp->pae_authsess_mab();
+  my $pae_capabilities = $snmp->pae_i_capabilities();
+  my $capdesc = { 'dot1xPaePortAuthCapable' => , 'Authenticator functions are supported',
+   'dot1xPaePortSuppCapable' => , 'Supplicant functions are supported'};
```
ollyg, line 42: So for LLDP we don't store the capabilities and instead have boolean fields for the capabilities. Would you consider doing the same here? (such as `pae_authenticator_capable` and `pae_supplicant_capable`)

I'm not so happy for long english text to be stored in the field, and that would be better done in the user interface in response to the setting in the database.

Also can a port have multiple PAE capabilities? Because the current code does not support it.

rc9000: Yes right it's actually a bit field, not an enum, so both could be true. Should work in the new commit.

lib/App/Netdisco/Util/PortAccessEntity.pm
```diff
@@ -0,0 +1,81 @@
...
+
+sub update_pae_attributes {
+
+  my ($device) = @_;
+  no warnings "uninitialized";
+
+  my $snmp = App::Netdisco::Transport::SNMP->reader_for($device)
+    or return Status->defer("pae failed: could not SNMP connect to $device");
+  my $interfaces = $snmp->interfaces;
+  debug sprintf ' [%s] pae - updating PortAccessEntity details for %s', $device->ip, $device->dns;
```
ollyg, line 32: devices might not have dns so perhaps just leave it as "updating pae details"

lib/App/Netdisco/Worker/Plugin/Discover/PortAccessEntity.pm
```diff
@@ -0,0 +1,18 @@
+package App::Netdisco::Worker::Plugin::Discover::PortAccessEntity;
+
+use Dancer ':syntax';
+use App::Netdisco::Worker::Plugin;
+use aliased 'App::Netdisco::Worker::Status';
+use Dancer::Plugin::DBIC 'schema';
+use App::Netdisco::Util::Worker;
+use App::Netdisco::Util::PortAccessEntity qw/update_pae_attributes/;
+
+register_worker({ phase => 'store', driver => 'snmp' }, sub {
```
ollyg, line 10: any reason this is phase 'store' and not phase 'main'?

lib/App/Netdisco/Worker/Plugin/Discover/PortProperties.pm
```diff
@@ -10,6 +10,7 @@ use Dancer::Plugin::DBIC 'schema';
 use Encode;
 use App::Netdisco::Util::FastResolver 'hostnames_resolve_async';
 use App::Netdisco::Util::Device qw/is_discoverable match_to_setting/;
+use App::Netdisco::Util::PortAccessEntity qw/update_pae_attributes/;
```
ollyg, line 13: don't know why this is here ... from an older version where it wasn't a separate plugin?

lib/App/Netdisco/Worker/Plugin/Discover/PortAccessEntity.pm
ollyg: ah I think I've worked it out ... you want PortAccessEntry to happen after PortProperties (which deletes rows). Try naming it `App::Netdisco::Worker::Plugin::Discover::PortProperties::PortAccessEntity`

rc9000: Yes exactly, also insert. I'll try this sub-package variant.

lib/App/Netdisco/Worker/Plugin/Discover/PortProperties.pm
rc9000: exactly, gone now

lib/App/Netdisco/Worker/Plugin/Discover/PortAccessEntity.pm
rc9000: Worked like a charm. I also did it for the Macsuck plugin for symmetry reasons

PR #941, opened 2022-11-08
Add NAC columns to UI

rc9000:
This is a first stab at adding various NAC... attributes to the port and device details.

TBD:

 * is the last eapol frame source helpful
 * the whole MAB business is based on the MAC being in hyphenatend AA-BB-CC-DD-EE format. This might be exclusively a Cisco thing, unfortunately I have no other devices to look at. Do we leave it in like that?
 * should we add the is_supplicant/is_authenticator columns too?

Other suggestions etc. of course welcome.


ollyg:
@rc9000 thanks! I'm happy to help if I have a device snapshot with the data that I can develop against.

My first thought is whether we can display the info more compactly in the web interface. So instead of many columns (which expands the sidebar list and makes the table wide) we have perhaps a single column displaying the data, with icons if bool.

ollyg:
For the MAC addr format we can run it though NetAddr::MAC and then it will also pick up the user's preferred display format.

Not sure if you mean the MAC addr is being stored in Cisco format (I thought postgressql normalised, or is it text field?).

rc9000:
Ok I realized that my points maybe don't make much sense without some data to look at :) Unfortunately I can't share a full snapshot, there is too much confidential customer data in these. 

Here's a screenshot what the table roughly looks like for an average Cisco with NAC enabled:

scr 2022-11-08 at 22 31 00

Same thing as sql update statements - with slapping on a WHERE and changing the IP on some lines you could update a random interfaces to have some NAC data:

[pae_sample.sql.txt](https://github.com/netdisco/netdisco/files/9965432/pae_sample.sql.txt)


I agree that it is a lot of columns, I was also thinking about just making a single "NAC Status" column with all the info combined. Then I went with individual columns because I thought it would be more useful when exporting CSV or when searching for stuff. But I'm still open to both, I'll make a single column variant to see how it would look.

ollyg:
@rc9000 what does it mean when State is 'disconnected' and Status is 'unauthorized' but there's still a user and/or mab and/or last eapol frame source.

Also, wonder if the mab column should've been a boolean.

rc9000:
@ollyg I've tried a compact "NAC Summary" column now. I also dropped the last eapol source completely, and MAB is not a separate column anymore but shown behind the "user" if it happens to be a MAC address in Radius format. 

Currently I've left the individual columns and the summary, but I think we could just leave summary and drop the others. 
Example how it looks like now:

scr 2022-11-28 at 22 18 18

Unfortunately reading this stuff is not trivial, examples:

 * Gi1/0/9 is a port that actually did dot1x and is happily authenticated
 * Gi1/0/8 is connected with MAB, as the port is up and the user is show as a MAC address how Cisco defines them for  lookup in Radius. You'd think it doesn't work because it says EAP State disconnected, but that refers to the EAP state machine defined in the protocol. 
 * The worst part is that attributes don't clear when the client is gone, e.g. Gi1/0/3 still fondly remembers the last client even if the port is down.

I was thinking of showing this stuff only when the port is up, but since we've discovered that only discover updates ifoper, that's maybe a bit premature.

What do you think about the NAC Summary variant? Once we've settled on the columns, I'll still need to do the CSV bits.

ollyg:
Hi @rc9000 could you please give the branch a spin and let me know what you think of the web summary now? Of course I had to develop a bit blind as I don't have .1X stuff running on my laptop at home ;)

Yes perhaps the icons are a bit cryptic but the full detail is in CSV if needed and I think user is mostly interested in overall success and the "user". Netdisco tries now to make a good presentation of the user and link to node search.

Thanks!

ollyg:
Actually I will do an update and add another option to the NAC display, to show the full column detail in the web, I've worked out how to make it happen without being too intrusive (similar to extended LLDP info we can display with extra second tickbox)

ollyg:
Done the update -- tooltips on the icons

rc9000:
@ollyg Sorry it took me a while but I finally got around to having a look. Works great and looks nice with the icons, cool! I'd say good to go.

PR #942, opened 2022-11-21
API Endpoints to submit arpnip and macsuck results

### LINKED ISSUE #893: FR - API Endpoints to submit arpnip and macsuck results
It would be great if we could submit arpnip and macsuck results via API. Usecases:

 * results from devices with poor SNMP support but other means to read arp and mac address tables (e.g Palo Alto in #755)
 * results from devices with broken VLAN indexing (e.g. Cisco IOS 17 in #883) 
 * importing flat files because people don't trust me with SSH on their equipment :) 

I think this could open up new classes of devices with significantly lower barrier to entry than we had with e.g. the Worker Plugin-based SSH collector.

ollyg:
resolves #893


ollyg:
Thanks @rc9000 !
I didn't see the issue you report myself, in testing (using same plackup command) but will have a poke and see what could be going on.

ollyg:
@rc9000 you could also try ?enqueue=true and see if it makes a difference (I wonder if running the backend job inline in the web process is breaking something - it was never intended to work that way!)

...which reminds me, that when using enqueue=true the implementation sets timestamp to when the job is run, not when you submit the data (which would be better), so I'll patch for that.

rc9000:
Oh nevermind it just doesn't work in netdisco-web-fg, the requests > 1 fail with job log:

          'error': 'failed to report from any worker!'

Probably because of the single process environment. When launching with `netdisco-web start` it works fine.

ollyg:
@rc9000 but that is strange because that's how I'm developing and testing ... in the foreground. I will have a closer look...

You can also use the command line, btw:
`netdisco-do macsuck -d x.x.x.x -p /path/to/file/with/the/data.json`

ollyg:
@rc9000 aha! I can reproduce, thank you

ollyg:
pushed updates to the PR which remove the enqueue option for web submit, and also fix a couple bugs

PR #961, opened 2023-01-10
Announce when we start updating stats

gavinatkinson:


ollyg:
Hi @gavinatkinson I'm not sure I understand what you want to achieve with the patch. The existing code shows "Updating statistics..." when the statistics update starts, and then "done." when it's complete. Your patch seems to put this on two lines instead of one, but no change to when the text appears. Can you help me understand?

Many thanks!

gavinatkinson:
Hi,

Apologies for the lack of details.  The issue is that updating the statistics can take many minutes, and (because STDOUT defaults to being line-buffered in perl?) the "Updating statistics... " part only gets printed after the statistics update finishes, and not before (as only then does LF get printed).  As a result the script just appears to hang after completing the DB schema update - in the case that caused me to look at this, the stats update took about 20 minutes.

It may be that a better solution is to set STDOUT to be unbuffered (with `$| = 1;` at the start of the code) - I've tested this and this also seems to work.

Gavin

ollyg:
20 minutes?! wow, OK. Certainly I'll apply the patch for the buffered IO but if you have the time and energy I'd appreciate finding the slow query. You could run with DBIC_TRACE=1 in the environment and see which query it pauses at.

ollyg:
thanks @gavinatkinson !
implemented in 824cf320

Do let me know about the slow query if you find it

PR #979, opened 2023-02-14
Added tenancy awareness to generic reports

dtuecks:
Custom reports did not support tenants. Please review if we can replace $report->{database} with $vars->{'tenant'}.


ollyg:
thanks @dtuecks !
Implemented alternative fix in 86ef8add

Issue #1016, opened 2023-04-03
SNMP IP addresses not correctly recorded for Fortigate with version 7.2.x


earendilfr:
## Expected Behavior

I try to discover a new firewall Fortigate with the firmware version 7.2.4.
But, when I discover it with netdisco, I cannot successfully record the IP address of the interfaces

## Current Behavior

When I do it with the CLI tool, I can see the IP addresses from the SNMP MIB but nothing is recorded in the addresses tab
I suspect that Fortinet has changed the IP-MIB and now, the key is not the IP address but the concatenation of the IP address and the ID of the VDOM where is attached the interface.

You can found bellow the result of a SNMP walk for a fortigate with vervsion 7.2.4:
```
IP-MIB::ipAdEntIfIndex.10.1.0.1.3 = INTEGER: 38
IP-MIB::ipAdEntIfIndex.10.2.0.1.3 = INTEGER: 39
IP-MIB::ipAdEntIfIndex.127.0.0.1.3 = INTEGER: 45
IP-MIB::ipAdEntIfIndex.10.0.0.240.1 = INTEGER: 2
IP-MIB::ipAdEntIfIndex.10.0.0.241.1 = INTEGER: 2
```

and the result on a fortigate with the version 6.4.9
```
IP-MIB::ipAdEntIfIndex.10.0.0.243 = INTEGER: 2
IP-MIB::ipAdEntIfIndex.127.0.0.1 = INTEGER: 137
IP-MIB::ipAdEntIfIndex.192.0.0.209 = INTEGER: 136
IP-MIB::ipAdEntIfIndex.192.1.0.35 = INTEGER: 137
```

Both firewall are configured with VDOM

## Possible Solution

I have a proposal on code part `perl5/lib/perl5/App/Netdisco/Worker/Plugin/Discover/Properties.pm` with the use of ip_table to retrieve the real IP address of an interface.
But I don't know if you want to solve this issue in netdisco or in SNMP::Info in code related to Fortinet (because it's seems to be a change done by fortinet to indicate the VDOM related to an interface).
If you are welcome to solve this issue in Netdisco code, I will propose a PR to solve this issue.

## Your Environment

* Netdisco version used: 2.60.10
* SNMP::Info version used: 3.92

## Device information

```
$ netdisco-do show -d 10.208.233.241 -e specify -DI
[1569638] 2023-04-03 22:25:52  info App::Netdisco version 2.060010 loaded.
[...]

PR #1026, opened 2023-04-11
Change the way to retrieve IP address of interfaces

earendilfr:
Use by default the ipTable MIB entry to retrieve the IP address of interfaces and fallback to the index method as previously. 
Fixe the issue with the case where the index is not really an IP address (seen in Fortinet with VDOM).

Related to issue #1016


ollyg:
Many thanks for the patch @earendilfr now merged!

PR #1045, opened 2023-06-08
Add indexes for custom_fields in device + device_port

rc9000:
Hello @ollyg & team. We've put quite a bunch of additional attributes into custom_fields. Works great but got a bit slow, could we add these indexes to speed up various jsonb operations?

This requires GIN and the jsonb_ops, AFAICT this should work in 9.6 already: https://www.postgresql.org/docs/9.6/gin-builtin-opclasses.html - but I've only tried it in >= 12


ollyg:
thanks!

PR #1046, opened 2023-06-08
Try to fix #1044 by subquery inlining instead of CTE

### LINKED ISSUE #1044: PortVLANMismatch SQL materialises the WITH clause
Virtual::PortVLANMismatch uses a chunky SQL query with a WITH clause.

It looks like adding "NOT MATERIALIZED" to this clause gives a ~ 30x speedup, enough to prevent the report timing out in the web UI in large environments.

rc9000:


ollyg:
thanks, this is a great start, and I might optimise

PR #1096, opened 2023-08-18
App::Netdisco::SSHCollector::Platform::FortiOS

earendilfr:
Update the SSH Colector for FortiOS to support the multivdom environment

By default, currently, the module think you didn't have VDOM.
With this change, the module search if you are in a VDOM context and, in this case, perform the lookup on all the context

### DIFF
--- a/lib/App/Netdisco/SSHCollector/Platform/FortiOS.pm
+++ b/lib/App/Netdisco/SSHCollector/Platform/FortiOS.pm
@@ -16,6 +16,7 @@ use warnings;
 use Dancer ':script';
 use Moo;
 use Expect;
+use Regexp::Common qw(net);
 
 =head1 PUBLIC METHODS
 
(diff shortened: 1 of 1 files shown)


rc9000:
Check out what I just did earlier this week :) https://github.com/netdisco/netdisco/pull/1092 - what a coincidence. 

What I like better in your version is using Regexp::Common and using the "get system" commands to find the VDOM info. Somehow my user was not allowed to do this, so I tediously parsed it from "show" which seems quite fragile. 

Do you know what exactly is needed in the [accprofile](https://docs.fortinet.com/document/fortigate/7.4.0/cli-reference/2620/config-system-accprofile) to run the get system and diagnose commands?

earendilfr:
Thanks a lot 😄 

> Do you know what exactly is needed in the [accprofile](https://docs.fortinet.com/document/fortigate/7.4.0/cli-reference/2620/config-system-accprofile) to run the get system and diagnose commands?

I'm not because, in my case, I have used the same login/password used for rancid to do backup...
But I have tested with a user that have a global `read-only` access and it's working fine.

PR #1104, opened 2023-09-26
Open URL on statisticcs page in a new tab

earendilfr:
Just a cosmetic change: Update the statistics template to add an option to open external URL in a new tab


ollyg:
thank you @earendilfr !

PR #1117, opened 2023-11-03
FortiOS.pm improvements

rc9000:
@earendilfr & @ollyg - this is a slightly improved version of the FortiOS collector

 * can handle --More-- pagination
 * suppress some uninitialized variable warnings

### DIFF
--- a/lib/App/Netdisco/SSHCollector/Platform/FortiOS.pm
+++ b/lib/App/Netdisco/SSHCollector/Platform/FortiOS.pm
@@ -50,18 +50,18 @@ sub get_paginated_output {
         if ($match) {
             if ($match =~ $more_pattern) {
                 $more_flag = 1;
-		        @lines = split(/\R/, $before);
-		        push(@alllines, grep {$_ =~ /\S/} @lines);
-		        debug("skipping through --More-- pagination");
+                @lines = split(/\R/, $before);
+                push(@alllines, grep {$_ =~ /\S/} @lines);
+                debug("skipping through --More-- pagination");
                 $expect->send(" ");
             } elsif ($match =~ $prompt) {
                 $more_flag = 0;
-		        @lines = split(/\R/, $before);
-		        push(@alllines, grep {$_ =~ /\S/} @lines);
-		        foreach my $line (@alllines) {
-		            debug("output collected: $line") if $line;
-		        }
-		        last;
+                @lines = split(/\R/, $before);
+                push(@alllines, grep {$_ =~ /\S/} @lines);
+                foreach my $line (@alllines) {
+                    debug("output collected: $line") if $line;
+                }
+                last;
             }
         }
     }
@@ -142,7 +142,7 @@ sub arpnip {
     foreach (@data) {
         if ($_ && /^Virtual domain configuration: multiple$/) {
             $multi_vdom = 1;
-	    last;
+        last;
         }
     }
     my $arpentries = [];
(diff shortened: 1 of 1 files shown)


earendilfr:
Hi @rc9000,

Two question: 
- how did you have the pagination for the output of `get system arp`? It's seems not available (even with a Fortigate with 1756 arp entries...) Or perhaps it's for IPv6 entry (sadly, I didn't have IPv6 on my firewalls,,,)
- Are you sure it's a real good idea to declare variable `$prompt` and `$more_prompt` as òur`variables... I read it's a better wait to declare them as my (that will cache this variable from outside of the module)

Some other points:
Your idea to declare regex outside and compile then is really a good idea.
I think you can also add two other variables
```perl
my $re_ipv4_arp = qr/^($RE{net}{IPv4})\s*\d+\s*($RE{net}{MAC})\s*\S+$/;
my $re_ipv6_arp = qr/^ifindex=\d+\s+ifname=\S+\s+($RE{net}{IPv6}{-sep => ':'}{-style => 'HeX'})\s+($RE{net}{MAC}).*$/;
```
and use them inside the function `arpnip_context`

And perhaps you can simplify the function `sub get_paginated_output` like that:

```perl
sub get_paginated_output {
    my ($command, $expect) = @_;
    $expect->send($command."\n");
    my ($pos, $error, $match, $before, $after) = $expect->expect($timeout,
        [ $more_pattern, sub { my $self = shift; $self->send(" "); exp_continue; }],
        -re, $prompt);
    return grep {$_ =~ /\S/} split(/\R/, $before);
}
```

rc9000:
Thanks for having a look @earendilfr !

 
>     * how did you have the pagination for the output of `get system arp`? It's seems not available (even with a Fortigate with 1756 arp entries...) Or perhaps it's for IPv6 entry (sadly, I didn't have IPv6 on my firewalls,,,)

Hehe weird you're right, I actually just noticed the problem in "get system status" and just assumed that the other outputs also would have paging - but "get system arp" isn't doing it for me neither. I also don't know about IPv6, because the admins didn't want to give my netdisco user the `diagnose` command. But I'm curious now, if I encounter some IPv6 Fortis somewhere I'll give it a try :)



>     * Are you sure it's a real good idea to declare variable `$prompt` and `$more_prompt` as òur`variables... I read it's a better wait to declare them as my (that will cache this variable from outside of the module)

Totally true, my is sufficient. 

> Some other points: Your idea to declare regex outside and compile then is really a good idea. I think you can also add two other variables
> 
> ```perl
> my $re_ipv4_arp = qr/^($RE{net}{IPv4})\s*\d+\s*($RE{net}{MAC})\s*\S+$/;
> my $re_ipv6_arp = qr/^ifindex=\d+\s+ifname=\S+\s+($RE{net}{IPv6}{-sep => ':'}{-style => 'HeX'})\s+($RE{net}{MAC}).*$/;
> ```

added, good point. 


> And perhaps you can simplify the function `sub get_paginated_output` like that:
> 
> ```perl
> sub get_paginated_output {
>     my ($command, $expect) = @_;
>     $expect->send($command."\n");
>     my ($pos, $error, $match, $before, $after) = $expect->expect($timeout,
>         [ $more_pattern, sub { my $self = shift; $self->send(" "); exp_continue; }],
>         -re, $prompt);
>     return grep {$_ =~ /\S/} split(/\R/, $before);
> }
> ```

Yes this should probably work, but I liked the explicitness and logging of the current variant, so I left it at that.

PR #1121, opened 2023-11-10
Fix a potential error with undefined value in Transport/SNMP.pm

earendilfr:
Sometimes, I can see in the job queue that some jobs are in a wrong format because we try to call a new method to an undefined object
image

I suspect that for an unknown reason, sometimes, the $class variable is undefined...

So, I have added a test to validate that the object is defined prior to call the new method

### DIFF
--- a/lib/App/Netdisco/Transport/SNMP.pm
+++ b/lib/App/Netdisco/Transport/SNMP.pm
@@ -216,11 +216,13 @@ sub _snmp_connect_generic {
           # if successful, restore the default/user timeouts and return
           if ($info) {
               my $class = ($useclass ? $classes[0] : $info->device_type);
-              return $class->new(
-                %snmp_args, Version => $ver,
-                ($info->offline ? (Cache => $info->cache) : ()),
-                _mk_info_commargs($comm),
-              );
+              if (defined $class && eval { $class->can("new") }) {
+                  return $class->new(
+                    %snmp_args, Version => $ver,
+                    ($info->offline ? (Cache => $info->cache) : ()),
+                    _mk_info_commargs($comm),
+                  );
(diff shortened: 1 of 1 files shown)


ollyg:
thanks @earendilfr !
the reason I didn't yet merge this is I'd like to understand why it's happening, as it should not, and maybe the fix is to refactor instead of workaround. thanks for raising the issue, I will take a look.
if you can somehow work out any pattern or reproducing device, do let us know!

earendilfr:
Hi @ollyg, I will try to reproduce the reason of this error.
One difficulty is the error are not showed in netdisco-backend logs, only in jobqueue through WebUI...

inphobia:
lemme know what you need, always have some of these in my logs

image

see this with discover jobs most of the time, in rare cases with arpnip.  macsuck i can't see in my current list.

also get this at times, just no message:

image

one very important note, i'm using:

```
# discover everything without serials
snmp_field_protection:
  device:
    serial: ['!group:__ANY__']
```

the ip range tells me this is a 4g device using nbm (network behind mobile. using private apn & provider pushes the routes when it comes online).  we have around 2000 such devices & our field team swaps them, often with a different vendor depending on what we have in stock.

earendilfr:
I confirm I have also tasks in error without any reason when I check:
image

But I didn't have the config part indicated by @inphobia.

Perhaps an help concerning this incident could be to add, in job queue log, the date and hours where the job was started and finished...

ollyg:
I'm going to close this longstanding patch, as I don't really like just hiding the error, and it's not clear why the error is happening or that it's materially impacting netdisco. happy to open a new patch with a fix if it comes. just housekeeping.

PR #1139, opened 2023-12-13
Fix an issue when you try to search a false MAC

earendilfr:
Currently, when you search a false MAC address like `e2:62:a3:31:5e:734` or `ae2:62:a3:31:5e:73`, you receive an error 500.
So, a simple fix to filter correctly when the MAC address is bad.

But, I think a better solution should be send back an error message in WebUI to indicate that the MAC address is wrong...


ollyg:
many thanks!

Issue #1053, opened 2023-06-23
FR: store origin of node_ip entries


rc9000:
IRC "specification":

```
12:07 < Gooses> but it won't know about the IP until it's repolled the router for that network
12:08 < Gooses> if there was a 'refresh routers' button that'd be super handy
12:08 < Gooses> repolling everything is an option but the poor server is a bit overloaded
12:16 < oliver> Gooses: yes isn't that running an arpwalk ?
12:17 < oliver> presumably you're doing a macsuck on the device to see the node appear
12:18 < Gooses> yes, but the IP address doesn't show until I invoke an all job
12:18 < Gooses> I assume that just refreshing the device doing routing for that network would e enough
12:45 < oliver> Gooses: yes but "refresh" is an imprecise term. You have discover, macsuck, arpnip as the three actions. For the local switch with the new node, a macsuck should be sufficient, and for the router, an arpnip
                should be sufficient. macwalk and arpwalk would also do the same of course.
12:46 < oliver> (and in the default schedule they are happening every hour)



12:49 < rc9000> I would love for node_ip to record the source information where we found the arp entry, it would make re-arpnipping for diagnostics etc. a lot easier. especially in a dc environment
                where "what is the router" seems to be a non-trivial question.

13:53 < oliver> ... simple as adding column to node_ip and adding the device IP there and either updating PK or creating an index on it ?

10:51 < oliver> rc9000: I wonder if the "seen by" device we discussed adding to node_ip could be an array of inet field, rather than adding new rows and changing the PK. thoughts?
10:52 < oliver> oh this might be annoying for archive, so perhaps a non-starter, but would appreciate opinion anyway, just in case it turns out to work well
10:53 < oliver> one downside is not having timestamps for that seen_by event per device (there's still time_last on the row)
12:22 < rc9000> oliver: yes i was considering that, something like a FIFO with the last couple of records. could even be a jsonb array of objects with the hostname and the date. then we wouldn't need to
                fix all of the reports that break etc.
12:24 < rc9000> and the archive issue can be ignored, if there's an ip in there that's no longer in device, so be it.
```

ollyg:
remember to update the API entry and API doc for arps as well

PR #1174, opened 2024-02-22
store origin device of node_ip entries in custom_fields.seen

rc9000:
Hello @ollyg & team! Here's a first version of what's discussed in #1053 and on the IRC.
If we agree on this implementation I'll make a second patch that shows this information in the UI.

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -66,8 +66,8 @@ Module::Build->new(
     'Net::DNS' => '0.72',
     'Net::LDAP' => '0',
     'Net::OpenSSH' => '0',
-    'Net::Ping' => '0', # is core
-    'Net::Ping::External' => '0',
+    'Net::Ping' => '2.75', # is core
+    'Net::Ping::External' => '0.15',
     'NetAddr::MAC' => '0.93',
     'NetAddr::IP' => '4.068',
     'Opcode' => '1.07',
--- a/Build.PL
+++ b/Build.PL
@@ -66,6 +66,8 @@ Module::Build->new(
     'Net::DNS' => '0.72',
     'Net::LDAP' => '0',
     'Net::OpenSSH' => '0',
+    'Net::Ping' => '0', # is core
+    'Net::Ping::External' => '0',
     'NetAddr::MAC' => '0.93',
     'NetAddr::IP' => '4.068',
     'Opcode' => '1.07',
--- a/Build.PL
+++ b/Build.PL
@@ -88,7 +88,7 @@ Module::Build->new(
     'Starman' => '0.4008',
     'Storable' => '0',
     'Sys::SigAction' => '0',
-    'SNMP::Info' => '3.95',
+    'SNMP::Info' => '3.970001',
     'SQL::Abstract' => '1.85',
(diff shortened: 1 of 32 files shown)


ollyg:
Good approach! My initial thoughts are below. These are just thoughts and not requirement to do anything different, and some is just thinking out loud the options. Let me know what you think.
- do we want to stomp on the "seen" slot of user-defined custom fields? we'd have to document it and guard against a user setting/messing with it. options are to use something other than custom_fields (like system_fields) or else to prefix our fields with netdisco_ or similar and document/protect that (easier?).
- the implementation makes a dictionary and updates the router entry with the timestamp. just noting that this is not a "history" which would be a list where entries are added when the router IP changes, or latest is updated (so you can go routerA, routerB, routerA again, etc). However as Netdisco doesn't support L3 virtualisation I think your implementation is safer, otherwise we could have an explosion if two routers flip-flop entries.
- I'd like to find a more efficient database update algorithm than update_or_new then insert with the branch, but it's not essential.

rc9000:
> 
>     * do we want to stomp on the "seen" slot of user-defined custom fields? we'd have to document it and guard against a user setting/messing with it. options are to use something other than custom_fields (like system_fields) or else to prefix our fields with netdisco_ or similar and document/protect that (easier?).

I like the prefix idea and suspect similar cases might come up in other tables, let's just say netdisco_ keys are  implementation namespace in all custom_fields columns and should not be used? Or maybe better `_netdisco_` since this is a convention people might be familiar with already.

> 
>     * the implementation makes a dictionary and updates the router entry with the timestamp. just noting that this is not a "history" which would be a list where entries are added when the router IP changes, or latest is updated 

True. It's nice that it's currently a self-managed structure that will not grow infinitely in size. Maybe as a one-step inbetween compromise we could have first/last fields like all other historic data does. 

> 
>     * I'd like to find a more efficient database update algorithm than update_or_new then insert with the branch, but it's not essential.

Yes that bugged me too and took basically all of the time that went into this PR, but I didn't find a better way :)

rc9000:
ok it's now `_netdisco_seen_in_arp` and with time_first/time_last. But updating deep json structures like this is pretty horrid with the jsonb_ functions. I gave up around

```
    $row->set_column(
        custom_fields => \[ qq{ jsonb_set( custom_fields,
                    '{_netdisco_seen_in_arp, ?}',
                    jsonb_build_object( ?::text, coalesce( custom_fields#>>'{_netdisco_seen_in_arp, ?}'::text[],
                            'null')::jsonb || jsonb_build_object( 'time_last', to_jsonb(to_timestamp(?))
                        )), true
                )
            },$device_ip, $device_ip, $device_ip, $now ]
    ) if $device_ip;
```

which also would have needed a similar jsonb_build_object counterpart for the insert case, and switched to JSON::XS `(en|de)code_json`. 

Roundtripping and overwriting the whole jsonb each time is probably less efficient, but way more readable. Shouldn't also get too big unless you run into #1185 which gave me millions of `_netdisco_seen_in_arp`in no time. I started to doubt the network design :)

Also added an index for the jsonb like all the other tables have.

Issue #1170, opened 2024-02-19
[Feature Request] Ping first, scan later


hac3ru:
Hello,

I would like to suggest / ask for a feature: sometimes, due to different reasons (you can read bad planning), you got a huge network with just a handful of devices online. On these networks, I found out that macsuck and arpnip will go through the entire network and try to scan each IP address. It would be amazing if we could have a config option / checkbox saying "Scan only hosts that reply to icmp". This way, even if we'd have a /16 network, with a handful of hosts, scanning it all would take minutes, since it would first icmp each IP and, if it doesn't reply, it will go to the next one, without further action on the unresponsive IP.

Hope I made that clear enough....

ollyg:
Hi @hac3ru yes that's clear. In the past I've seen people write a small shell script to do the ping and then queue a discover if the device responds. Your idea could be that netdisco has a job like discoverall which does the ping sweep and queues a discover for successful ones. I like it.

hac3ru:
From my point of view it would make for a more easy to use software. It would become quite friendly with non-scripters/programmers.

hac3ru:
P.S. coming back with an improvement, or so I believe: it would be best if this would be on a per subnet basis. Like a checkbox near "Discover". This way, we could have both the "regular" behavior and the ping before scan, depending on the network setup.

ollyg:
note to self: job should set high timeout, as it could be still like 20 mins to ping scan a /16 depending on how asynchronous the implementation is

ollyg:
> fping has been created for batch use and take network CIDR in argument :)  for nmap I can imagine 2 differents ways of using it : 1/ nmap -sP X.X.X.0/24 (for doing Iregular CMP requests) 2/ nmap -sS X.X.X.0/24 (for checking ACK answer only which should be quicker than a full ICMP request)
> 14:59 
> (typo : for doing regular ICMP requests)

inphobia:
net::ping requires root privs according to the docs (for icmp echo - use tcp by default), and spawning an external ping seems very costly performance wise.
i like the idea, but could be a challenge to implement in a portable & secure way.

inphobia:
perhaps an alternative could be to do this outside of netdisco: provide an example of how to use nmap/fping to find active devices, dump those to a file & use this as input for netdisco-do.

not quite sure why netdisco would queue up macsuck & arpnip for undiscovered devices as the original post mentioned.

hac3ru:
[...]

PR #1184, opened 2024-03-14
ping sweep utility worker to enqueue discover jobs from an IP prefix

ollyg:
This worker is a utility which runs ping against an IP prefix and enqueues a discover job for any host that responds.

You'll need to install Net::Ping::External and have the ping utility installed on your system.

IP prefix is passed in the -p (port) parameter.
Timeout is 0.1 seconds default or pass any number to the -e (extra) parameter.

The IPs are pinged sequentially so it's as slow as num hosts x timeout. To parallelise and make things faster, queue the pingsweep job and use netdisco workers.

You could also run this as additional schedule config periodically for one or more IP prefixes.

The timeout is achieved using a signal because the timeout param to CLI ping is either integer (too slow) or root privilege or simply missing on some *nix platforms.

### DIFF
--- a/Build.PL
+++ b/Build.PL
@@ -66,6 +66,8 @@ Module::Build->new(
     'Net::DNS' => '0.72',
     'Net::LDAP' => '0',
     'Net::OpenSSH' => '0',
+    'Net::Ping' => '0', # is core
+    'Net::Ping::External' => '0',
     'NetAddr::MAC' => '0.93',
     'NetAddr::IP' => '4.068',
     'Opcode' => '1.07',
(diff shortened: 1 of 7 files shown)


ollyg:
implements #1170

ollyg:
Might be better to queue as the hosts respond rather than all at the end. More efficient for large prefixes as get going sooner. 

Also note to change log output at end. 

Also note to find a way to use -d instead of -p if I can.

ollyg:
A /16 will be just under two hours to scan, I reckon.

ollyg:
Need to set two hour timeout

ollyg:
Hi @FabioPedretti not needed, as it's in the Build.PL file which lists dependencies, so if the user follows install instructions they will get the dependencies installed.

PR #1190, opened 2024-03-20
Implement of a MAC collector through SSH

earendilfr:
Some Cisco devices doesn't support the `macsuck` function because they doesn't support the BridgeMIB to the retrieve MAC addresses connected to switch ports...

(By example, the Cisco C1100 series routers have a switch inside but doesn't support the BridgeMIB...)

So, I have created a simple macsuck function through the SSH collector to retrieve these data.
Like the SNMP collector, the function:
- retrieve the information concerning the current statu of interface
- retrieve the MAC table

Currently, I have done only the IOS platform but it could be interesting for other devices...

### DIFF
--- a/lib/App/Netdisco/SSHCollector/Platform/IOS.pm
+++ b/lib/App/Netdisco/SSHCollector/Platform/IOS.pm
@@ -14,6 +14,7 @@ use strict;
 use warnings;
 
 use Dancer ':script';
+use NetAddr::MAC qw/mac_as_ieee/;
 use Moo;
 
 =head1 PUBLIC METHODS
@@ -30,6 +31,21 @@ Returns a list of hashrefs in the format C<{ mac =E<gt> MACADDR, ip =E<gt> IPADD
 =back
 
 =cut
+
+my $if_name_map = {
+  Vl => "Vlan",
+  Lo => "Loopback",
+  Fa => "FastEthernet",
+  Gi => "GigabitEthernet",
+  Tw => "TwoGigabitEthernet",
+  Fi => "FiveGigabitEthernet",
+  Te => "TenGigabitEthernet",
+  Twe => "TwentyFiveGigE",
+  Fo => "FortyGigabitEthernet",
+  Hu => "HundredGigE",
+  Po => "Port-channel",
+  Bl => "Bluetooth",
+};
 
 sub arpnip {
     my ($self, $hostlabel, $ssh, $args) = @_;
@@ -49,6 +65,42 @@ sub arpnip {
     }
 
     return @arpentries;
+}
+
+sub macsuck {
+  my ($self, $hostlabel, $ssh, $args) = @_;
+  
+  debug "$hostlabel $$ macsuck()";
+  my $cmds = <<EOF;
+terminal length 0
+show interfaces description
+show mac address-table
+EOF
+  my @data = $ssh->capture({stdin_data => $cmds}); chomp @data;
+  if ($ssh->error) {
+    info "$hostlabel $$ error in SSH command " . $ssh->error;
+    return;
+  }
+
+  my $interfaces = {};
+  my $macentries = {};
+  my $re_interface_line = qr/^([a-z]+)([0-9\/\.]+)\s+(up|down|admin down)\s+(up|down)/i;
+  my $re_mac_line = qr/^\s*(All|[0-9]+)\s+([0-9a-f]{4}\.[0-9a-f]{4}\.[0-9a-f]{4})\s+\S+\s+([a-zA-Z]+)([0-9\/\.]+)/i;
+  my $port;
+  foreach (@data) {
+    if ($_ && /$re_interface_line/) {
+      $port = sprintf "%s%s", $if_name_map->{$1}, $2;
+      $interfaces->{$port} = {
+        up_admin => ($3 eq "admin down") ? "down" : "up",
+        up => ($3 eq "down") ? "down" : "up",
+      };
+    } elsif ($_ && /$re_mac_line/) {
+      $port = sprintf "%s%s", $if_name_map->{$3}, $4;
+      ++$macentries->{$1}->{$port}->{mac_as_ieee($2)};
+    }
+  }
+
+  return ($interfaces,$macentries);
 }
 
 1;
--- a/lib/App/Netdisco/Transport/SSH.pm
+++ b/lib/App/Netdisco/Transport/SSH.pm
@@ -52,6 +52,11 @@ Returns C<undef> if the connection fails.
     my $self = shift;
     $self->platform->arpnip(@_, $self->host, $self->ssh, $self->auth);
   }
+
+  sub macsuck {
+    my $self = shift;
+    $self->plateform->macsuck(@_, $self->host, $self->ssh, $self->auth);
+  }
 }
 
 sub session_for {
--- a/lib/App/Netdisco/Worker/Plugin/Macsuck/Nodes.pm
+++ b/lib/App/Netdisco/Worker/Plugin/Macsuck/Nodes.pm
@@ -112,6 +112,36 @@ register_worker({ phase => 'main', driver => 'direct',
   return Status->done("Received MAC addresses for $device");
 });
 
+reister_worker({ phase => 'main', driver => 'cli',
+  title => 'gather macs from CLI and set interfaces'}, sub {
+
+  my ($job, $workerconf) = @_;
+  
+  my $device = $job->device;
+  my $cli = App::Netdisco::Transport::SSH->session_for($device)
+    or return Status->defer("macsuck failed: could not SSH connect to $device");
+
+  # Retrieve data through SSH connection
+  my ($interfaces,$macentries) = $cli->macsuck;
+
+  # make sure ports reflect their latest state as reported by device
+  foreach my $port (keys %{ vars->{'device_ports'} }) {
+    my $iif = $interfaces->{$port} or next;
+
+    debug sprintf ' [%s] macsuck - updating port %s status : %s/%s',
+      $device->ip, $port, ($iif->{up_admin} || '-'), ($iif->{up} || '-');
+
+    vars->{'device_ports'}->{$port}->update({
+      up => $iif->{up} || '-',
+      up_admin => $iif->{up_admin} || '-',
+    });
+  }
+
+  # get forwarding table and populate fwtable
+  vars->{'fwtable'} = $macentries
+
+  return Status->done("Gathered MAC addresses for $device");
+});
 
 register_worker({ phase => 'main', driver => 'snmp',
   title => 'gather macs from snmp and set interfaces'}, sub {


rc9000:
Cool! 👍  For the record, I had similar plans a while ago and already talked @ollyg into adding an API endpoint that can upload a macsuck result (`api/v1/object/device/{{hostname}}/nodes?enqueue=false`). I thought this would allow using ntc-templates instead of adding a lot of command parsing code in modules again. I did some experiments with that here: https://github.com/rc9000/ntcsuck/, it's bascially macsuck for IOS-XE via ntc-templates and Ansible.

However this approach showed two issues:

- `show mac address table` and similar does not seem to be a priority in ntc-templates and is poorly implemented for very common platforms even, so we'd probably need to supply our own TextFSM files anyways
- macsuck does a lot more than just submitting pure cam table results, a non-snmp macsuck will also have to deal with ifoperstatus etc. ... but I see you got that already covered to some extent. We'd probably need a way to still run the Wirelessnodes, NAC etc. stuff while not doing the cam/operatstatus parts. 

Then other stuff came inbetween and I parked the idea for the time being, looks like this was already two years ago now, dang. But I'd still like a way to do macsuck without snmp, my main problem is that it is so wasteful and slow on big switches and with community based indexing. And it doesn't even work right with SNMPv3 or is a pain with having to create a context per VLAN. 

This just for a bit of context, I'm in favor of both a ntc-templates or an expect-ish SSH variant. If we go with the latter, sorry @ollyg that I insisted on the API and then not really came through with any useful application for it.

ollyg:
@rc9000 don't feel bad! The work to support API actually means that inside Netdisco the gather of data is decoupled from the sanity checking and the storage phases, so it makes this work here easier to implement, and also generally cleaner design internally. I was very happy to do it. Note that we even support CLI input of MAC/ARP data via netdisco-do.

ollyg:
Ah yes, I see @earendilfr has implemented using exactly the right feature as a result of refactor! Makes me happy :-)
https://github.com/netdisco/netdisco/pull/1190/files#diff-2fe03540adb4f0ad088f140864852c8dd0b91b3a3e8990638137bf57010bc4bbR141

earendilfr:
> Cool! 👍 For the record, I had similar plans a while ago and already talked @ollyg into adding an API endpoint that can upload a macsuck result (`api/v1/object/device/{{hostname}}/nodes?enqueue=false`). I thought this would allow using ntc-templates instead of adding a lot of command parsing code in modules again. I did some experiments with that here: https://github.com/rc9000/ntcsuck/, it's bascially macsuck for IOS-XE via ntc-templates and Ansible.
> 
> However this approach showed two issues:
> 
>     * `show mac address table` and similar does not seem to be a priority in ntc-templates and is poorly implemented for very common platforms even, so we'd probably need to supply our own TextFSM files anyways
> 
>     * macsuck does a lot more than just submitting pure cam table results, a non-snmp macsuck will also have to deal with ifoperstatus etc. ... but I see you got that already covered to some extent. We'd probably need a way to still run the Wirelessnodes, NAC etc. stuff while not doing the cam/operstatus parts.
> 
> 
> Then other stuff came inbetween and I parked the idea for the time being, looks like this was already two years ago now, dang. But I'd still like a way to do macsuck without snmp, my main problem is that it is so wasteful and slow on big switches and with community based indexing. And it doesn't even work right with SNMPv3 or is a pain with having to create a context per VLAN.
> 
> This just for a bit of context, I'm in favor of both a ntc-templates or an expect-ish SSH variant. If we go with the latter, sorry @ollyg that I insisted on the API and then not really came through with any useful application for it.

You are right, the solution is not perfect (clearly, I prefer the SNMP way because the SSH is **very** slow).

But if you check the output, you can do the necessary to have in parallel the CLI and the SNMP (it's could be fun if we could run the main `macsuck` in CLI and the `macsuck::PortAccessEntity` in SNMP...) :
```
netdisco-do -D macsuck -d ehc-mor-ert01
[89980] 2024-03-21 20:16:07  info App::Netdisco version 2.072003 loaded.
[89980] 2024-03-21 20:16:07  info macsuck: [10.12.32.61] started at Thu Mar 21 21:16:07 2024
[89980] 2024-03-21 20:16:07 debug macsuck: running with timeout 600s
[89980] 2024-03-21 20:16:07 debug => running workers for phase: check
[89980] 2024-03-21 20:16:07 debug -> run worker check/1000000 "internal::backendfqdn"
[89980] 2024-03-21 20:16:07 debug -> run worker check/1000000 "internal::snmpfastdiscover"
[89980] 2024-03-21 20:16:07 debug running with configured SNMP timeouts
[89980] 2024-03-21 20:16:07 debug -> run worker check/0 "macsuck"
[89980] 2024-03-21 20:16:07 debug Macsuck is able to run.
[89980] 2024-03-21 20:16:07 debug => running workers for phase: early
[89980] 2024-03-21 20:16:07 debug -> run worker early/0 "prepare common data"
[89980] 2024-03-21 20:16:07 debug => running workers for phase: main
[89980] 2024-03-21 20:16:07 debug -> run worker main/1000000 "gather macs from file and set interfaces"
[89980] 2024-03-21 20:16:07 debug skip: fwtable data supplied by other source
[89980] 2024-03-21 20:16:07 debug -> run worker main/200 "gather macs from CLI and set interfaces"
[89980] 2024-03-21 20:16:07 debug cli session cache warm: [10.12.32.61]
[89980] 2024-03-21 20:16:08 debug 10.12.32.61 89980 macsuck()
[89980] 2024-03-21 20:16:09 debug -> run worker main/100 "gather macs from snmp and set interfaces"
[89980] 2024-03-21 20:16:09 debug skip: namespace passed at higher priority
[89980] 2024-03-21 20:16:09 debug -> run worker main/100 "macsuck::nodes::portaccessentity"
[89980] 2024-03-21 20:16:09 debug skip: namespace passed at higher priority
[89980] 2024-03-21 20:16:09 debug -> run worker main/100 "macsuck::wirelessnodes"
[89980] 2024-03-21 20:16:09 debug snmp reader cache warm: [10.12.32.61]
[89980] 2024-03-21 20:16:09 debug [10.12.32.61:161] try_connect with v: 3, t: 0.2, r: 0, class: SNMP::Info::Layer3::CiscoSwitch, comm: <hidden>
[89980] 2024-03-21 20:16:10 debug => running workers for phase: store
[89980] 2024-03-21 20:16:10 debug -> run worker store/0 "save macs to database"
[89980] 2024-03-21 20:16:10 debug  [10.12.32.61] macsuck 24:f2:7f:c7:5b:be - port GigabitEthernet0/1/0 has undiscovered neighbor 10.17.193.16
[89980] 2024-03-21 20:16:10 debug  [10.12.32.61] macsuck - port GigabitEthernet0/1/0 vlan 10 : 1 nodes
[89980] 2024-03-21 20:16:10 debug  [10.12.32.61] macsuck - stored 1 forwarding table entries
[89980] 2024-03-21 20:16:10 debug  [10.12.32.61] macsuck - removed 0 fwd table entries to archive
[89980] 2024-03-21 20:16:10 debug => running workers for phase: late
[89980] 2024-03-21 20:16:10 debug -> run worker late/0 "macsuck::hooks"
[89980] 2024-03-21 20:16:10 debug  [10.12.32.61] hooks - 0 queued
[89980] 2024-03-21 20:16:10  info macsuck: finished at Thu Mar 21 21:16:10 2024
[89980] 2024-03-21 20:16:10  info macsuck: status done: Ended macsuck for 10.12.32.61
```

One big difficulty (root cause is Cisco devices) is the need to launch all the cmd in **one time** because ssh session is dead  after the usage of capture function and so, the reuse of this function generate an error...

rc9000:
> One big difficulty (root cause is Cisco devices) is the need to launch all the cmd in **one time**

There was also various issues in [IOSXR.pm](https://github.com/netdisco/netdisco/commits/master/lib/App/Netdisco/SSHCollector/Platform/IOSXR.pm) over the years. The current version uses Expect instead of capture(), this seemed to work fine in the end.

ollyg:
I hesitate strongly to say this ... has Net::Appliance::Session been tried? Unfortunately I have no devices to test on, here.

rc9000:
> I hesitate strongly to say this ... has Net::Appliance::Session been tried? Unfortunately I have no devices to test on, here.

I can confirm that Net::Appliance::Session works well with pretty much everything that Cisco has sold in the last two decades :) IIRC in the first couple SSHCollector modules I didn't use it because the dependencies were a pain in our airgapped DC equipped with quite old RHEL servers, Expect was just there. 

Nowadays everybody seems to have Internet-straight-to-critical-infrastructure pipelines with Docker, Artifactory and DevopsThisAndThat, so that should not be an issue anymore.

ollyg:
This patch is fine. The reason I've not merged it yet is I wanted to see if there's a way to let Netdisco do the interfaces status as SNMP and cache that, then let the SSH mac-address table command run. Then combine the two at the end.

ollyg:
Actually @rc9000 I wonder why we implemented as `Discover/PortProperties/PortAccessEntity.pm` instead of `Discover/PortAccessEntity.pm` in #937 ... the latter would have allowed it still to run with SNMP I think.

earendilfr:
> > I hesitate strongly to say this ... has Net::Appliance::Session been tried? Unfortunately I have no devices to test on, here.
> 
> I can confirm that Net::Appliance::Session works well with pretty much everything that Cisco has sold in the last two decades :) IIRC in the first couple SSHCollector modules I didn't use it because the dependencies were a pain in our airgapped DC equipped with quite old RHEL servers, Expect was just there.
> 
> Nowadays everybody seems to have Internet-straight-to-critical-infrastructure pipelines with Docker, Artifactory and DevopsThisAndThat, so that should not be an issue anymore.

I can try to update the PR to replace the usage of Net::SSH module by Except or Net::Appliance::Session...
If I read correctly the "Net::Appliance::Session", the good point is to remove the need to detect correctly the prompt (I have do some work on Rancid and clearly, the detection of the prompt is a real nightmare 😨)

ollyg:
note to self @ollyg: remove macs on forbidden vlans (sanity_vlans)

lib/App/Netdisco/Transport/SSH.pm
```diff
@@ -52,6 +52,11 @@ Returns C<undef> if the connection fails.
     my $self = shift;
     $self->platform->arpnip(@_, $self->host, $self->ssh, $self->auth);
   }
+
+  sub macsuck {
+    my $self = shift;
+    $self->plateform->macsuck(@_, $self->host, $self->ssh, $self->auth);
```
ollyg, line 58: has this code been tested? there's a typo here

lib/App/Netdisco/SSHCollector/Platform/IOS.pm
```diff
@@ -51,4 +67,40 @@ sub arpnip {
...
+  foreach (@data) {
+    if ($_ && /$re_interface_line/) {
+      $port = sprintf "%s%s", $if_name_map->{$1}, $2;
+      $interfaces->{$port} = {
+        up_admin => ($3 eq "admin down") ? "down" : "up",
+        up => ($3 eq "down") ? "down" : "up",
+      };
+    } elsif ($_ && /$re_mac_line/) {
+      $port = sprintf "%s%s", $if_name_map->{$3}, $4;
+      ++$macentries->{$1}->{$port}->{mac_as_ieee($2)};
```
ollyg, line 99: does this mean that VLAN can be "All" ? Netdisco probably needs zero instead.

lib/App/Netdisco/SSHCollector/Platform/IOS.pm
```diff
@@ -51,4 +67,40 @@ sub arpnip {
...
+  my $port;
+  foreach (@data) {
+    if ($_ && /$re_interface_line/) {
+      $port = sprintf "%s%s", $if_name_map->{$1}, $2;
+      $interfaces->{$port} = {
+        up_admin => ($3 eq "admin down") ? "down" : "up",
+        up => ($3 eq "down") ? "down" : "up",
+      };
+    } elsif ($_ && /$re_mac_line/) {
+      $port = sprintf "%s%s", $if_name_map->{$3}, $4;
```
ollyg, line 98: need to either fail if if_name_map lookup fails, or fall back to the provided text (currently sprintf will croak on undef)

lib/App/Netdisco/Transport/SSH.pm
earendilfr: Currently, I use it for 2 devices on my test platform...
Which error message did you have ?
```
$ netdisco-do -D macsuck -d 10.12.32.61
[284503] 2024-04-15 19:56:53  info App::Netdisco version 2.072003 loaded.
[284503] 2024-04-15 19:56:53  info macsuck: [10.12.32.61] started at Mon Apr 15 21:56:53 2024
[284503] 2024-04-15 19:56:53 debug macsuck: running with timeout 600s
[284503] 2024-04-15 19:56:53 debug => running workers for phase: check
[284503] 2024-04-15 19:56:53 debug -> run worker check/1000000 "internal::backendfqdn"
[284503] 2024-04-15 19:56:53 debug -> run worker check/1000000 "internal::snmpfastdiscover"
[284503] 2024-04-15 19:56:53 debug running with configured SNMP timeouts
[284503] 2024-04-15 19:56:53 debug -> run worker check/0 "macsuck"
[284503] 2024-04-15 19:56:53 debug Macsuck is able to run.
[284503] 2024-04-15 19:56:53 debug => running workers for phase: early
[284503] 2024-04-15 19:56:53 debug -> run worker early/0 "prepare common data"
[284503] 2024-04-15 19:56:53 debug => running workers for phase: main
[284503] 2024-04-15 19:56:53 debug -> run worker main/1000000 "gather macs from file and set interfaces"
[284503] 2024-04-15 19:56:53 debug skip: fwtable data supplied by other source
[284503] 2024-04-15 19:56:53 debug -> run worker main/200 "gather macs from CLI and set interfaces"
[284503] 2024-04-15 19:56:53 debug cli session cache warm: [10.12.32.61]
[284503] 2024-04-15 19:56:54 debug 10.12.32.61 284503 macsuck()
[284503] 2024-04-15 19:56:56 debug -> run worker main/100 "gather macs from snmp and set interfaces"
[284503] 2024-04-15 19:56:56 debug skip: namespace passed at higher priority
[284503] 2024-04-15 19:56:56 debug -> run worker main/100 "macsuck::nodes::portaccessentity"
[284503] 2024-04-15 19:56:56 debug skip: namespace passed at higher priority
[284503] 2024-04-15 19:56:56 debug -> run worker main/100 "macsuck::wirelessnodes"
[284503] 2024-04-15 19:56:56 debug snmp reader cache warm: [10.12.32.61]
[284503] 2024-04-15 19:56:56 debug [10.12.32.61:161] try_connect with v: 3, t: 0.2, r: 0, class: SNMP::Info::Layer3::CiscoSwitch, comm: <hidden>
[284503] 2024-04-15 19:56:56 debug => running workers for phase: store
[284503] 2024-04-15 19:56:56 debug -> run worker store/0 "save macs to database"
[284503] 2024-04-15 19:56:56 debug  [10.12.32.61] macsuck 24:f2:7f:c7:5b:be - port GigabitEthernet0/1/0 has undiscovered neighbor 10.17.193.16
[284503] 2024-04-15 19:56:56 debug  [10.12.32.61] macsuck - port GigabitEthernet0/1/0 vlan 10 : 1 nodes
[284503] 2024-04-15 19:56:56 debug  [10.12.32.61] macsuck - stored 1 forwarding table entries
[284503] 2024-04-15 19:56:56 debug  [10.12.32.61] macsuck - removed 0 fwd table entries to archive
[284503] 2024-04-15 19:56:56 debug => running workers for phase: late
[284503] 2024-04-15 19:56:56 debug -> run worker late/0 "macsuck::hooks"
[284503] 2024-04-15 19:56:56 debug  [10.12.32.61] hooks - 0 queued
[284503] 2024-04-15 19:56:56  info macsuck: finished at Mon Apr 15 21:56:56 2024
[284503] 2024-04-15 19:56:56  info macsuck: status done: Ended macsuck for 10.12.32.61
```

lib/App/Netdisco/SSHCollector/Platform/IOS.pm
earendilfr: Currently, the output of `show mac address-table` indicate outputs for ALL...
```
ehc-mor-ert01#sh mac address-table
          Mac Address Table
-------------------------------------------

Vlan    Mac Address       Type        Ports
----    -----------       --------    -----
 All    0100.0ccc.cccc    STATIC      CPU
 All    0100.0ccc.cccd    STATIC      CPU
 All    0100.0ccc.ccce    STATIC      CPU
 All    0180.c200.0000    STATIC      CPU
 All    0180.c200.0001    STATIC      CPU
 All    0180.c200.0002    STATIC      CPU
 All    0180.c200.0003    STATIC      CPU
 All    0180.c200.0004    STATIC      CPU
 All    0180.c200.0005    STATIC      CPU
 All    0180.c200.0006    STATIC      CPU
 All    0180.c200.0007    STATIC      CPU
 All    0180.c200.0008    STATIC      CPU
 All    0180.c200.0009    STATIC      CPU
 All    0180.c200.000a    STATIC      CPU
 All    0180.c200.000b    STATIC      CPU
 All    0180.c200.000c    STATIC      CPU
 All    0180.c200.000d    STATIC      CPU
 All    0180.c200.000e    STATIC      CPU
 All    0180.c200.000f    STATIC      CPU
 All    0180.c200.0010    STATIC      CPU
  10    24f2.7fc7.5bbe    DYNAMIC     Gi0/1/0
  10    f04a.027c.5bf4    STATIC      CPU
```

I'm not sure that we need to keep these entries...
If they are not present in the SNMP output, I think a better idea should be to remove these MAC address of the records...

ollyg: Thanks for the example! very helpful. I'll leave "All" in but set it to VLAN 0, and later on in Netdisco there's sanity checking code on all MAC addresses so it'll be fine.

ollyg:
replaced with #1202

PR #1208, opened 2024-05-13
sshcollector::nxos: Add macsuck function

earendilfr:
Add a macsuck collector function to module sshcollector::nxos


ollyg:
@earendilfr many thanks!

PR #1222, opened 2024-06-12
use `| no-more` instead of `terminal length 0` on NXOS

### LINKED ISSUE #1221: CLI driver (SSH) on Nexus might not discover all arp entries
Hi, we ran into an issue when we configured collection via SSH on our Nexus (N77c7706) devices.

## Expected Behavior
When doing an arpnip on the devices all arp entries should be discovered.

## Current Behavior
In our environment only a few arp entries (21 of 3800) are discovered. 
There is no hint of what goes wrong in the regular debug output.

```
[209755] 2024-06-11 13:03:22 debug ⮕ worker Arpnip::Nodes p200
[209755] 2024-06-11 13:03:22 debug cli session cache warm: [xxx.xxx.xxx.xx]
[209755] 2024-06-11 13:03:26 debug xxx.xxx.xxx.xx 209755 arpnip()
[209755] 2024-06-11 13:03:46 debug ⬅ (done) Gathered arp caches from xxx.xxx.xxx.xx
```
After turning on debug of the Expect module in `SSHCollector/Platform/NXOS.pm`
the error appears to be with switching off the terminal paging:
```
`terminal length 0\r\r\n% Permission denied for the role\r\n\r
```
then only one page (of 21 entries) is read from the arp table, output ends with
```
--More--
```

## Possible Solution
We wanted to give the user as few rights as possible, so we decided to add  `| no-more` 
to the two expect calls in `SSHCollector/Platform/NXOS.pm`.
In my understanding setting `terminal length 0` would be unnecessary then and this solution works even if the user has no rights to change the terminal settings.

`$expect->send("show ip arp vrf all | inc : | no-more\n");`

`$expect->send("show ipv6 neighbor vrf all | exclude Flags: | no-more\n");`

I would also be happy to provide a PR.

## Your Environment

* Netdisco version used:  	2.76.5
* SNMP::Info version used: 3.970.1

## Config info (deployment.yml)
```
- tag: nexusssh
    driver: cli
    platform: NXOS
    only: 
       - 'model:N77c7706'
    username: xxx
    password: yyy
```

j-hoff:
If the cli user is set up with the predefined priviledge level 1, the `terminal length 0` command is not allowed.
We can append a `| no-more` to the commands instead which has the same effect and does not need specific rights.

fixes #1221

### DIFF
--- a/lib/App/Netdisco/SSHCollector/Platform/NXOS.pm
+++ b/lib/App/Netdisco/SSHCollector/Platform/NXOS.pm
@@ -64,12 +64,9 @@ sub arpnip {
 
     ($pos, $error, $match, $before, $after) = $expect->expect($timeout, -re, $prompt);
 
-    $expect->send("terminal length 0\n");
-    ($pos, $error, $match, $before, $after) = $expect->expect($timeout, -re, $prompt);
-
     # we filter on the : in Age as the output header of the command may contain prompt chars, e.g.
     # Flags:   # - Adjacencies Throttled for Glean
-    $expect->send("show ip arp vrf all | inc :\n");
+    $expect->send("show ip arp vrf all | inc : | no-more\n");
     ($pos, $error, $match, $before, $after) = $expect->expect($timeout, -re, $prompt);
 
     my @arpentries;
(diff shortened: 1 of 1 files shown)


ollyg:
many thanks!

PR #1310, opened 2025-03-15
trying to fix validate/trust option combo

rc9000:
# Mailing list issue (by Michael, 11.03.)

We are working to set up SAML login via Apache Reverse Proxy to do authentication. We have the X-REMOTE_USER being populated with the user name from Entra ID. We have verified that the X-REMOTE_USER value is being sent to Dancer in the header. We did use the Dancer Debug sidebar to verify that the value is accurate. We do not see any login attempts into the Netdisco site from the user. In the deployment.yml file we have the values set as:

auth_user_env: "X-REMOTE_USER"
trust_x_remote_user: true
validate_remote_user: true

We have verified that the user exists in Netdisco but never see any authentication attempts come in. Is there additional configuration that needs to be done on the deployment.yml side to allow the user to authenticate? 

# My two cents

Yes these options don't seem to interact correctly. Attached is a quick AI-level fix that looks ok to me and seems to work, but there is a lot of interaction with all the various trust/validate/no_auth options, so I'd really appreciate if somebody can look at it in more detail too.

For testing I added an Apache RP with a static X-REMOTE_USER to our regular docker setup, and it also mounts DBIC.pm into so you can easily play around with it. I put the whole folder here: https://networkz.ch/tmp/remote-user-docker-compose.tar.gz (too big for attachment since I took the liberty to leave all the pg/log files in there).

### DIFF
--- a/lib/App/Netdisco/Web/Auth/Provider/DBIC.pm
+++ b/lib/App/Netdisco/Web/Auth/Provider/DBIC.pm
@@ -30,29 +30,38 @@ sub authenticate_user {
 
 sub get_user_details {
     my ($self, $username) = @_;
+    return unless defined $username;
 
     my $settings = $self->realm_settings;
     my $database = schema($settings->{schema_name})
         or die "No database connection";
 
-    my $users_table     = $settings->{users_resultset}       || 'User';
+    my $users_table     = $settings->{users_resultset}      || 'User';
     my $username_column = $settings->{users_username_column} || 'username';
 
+    # first try to find the user in the database
     my $user = try {
       $database->resultset($users_table)->find({
           $username_column => { -ilike => quotemeta($username) },
       });
     };
 
-    # each of these settings permits no user in the database
-    # so create a pseudo user entry instead
-    if (not $user and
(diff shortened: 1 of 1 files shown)


rc9000:
Ok since everybody on the mailing list seems to struggle with having to create the users and wanting to set permissons from the proxy, I did some more playing around. Now the PR supports new config, e.g.

# deployment.yml
    trust_x_remote_user: true
    validate_remote_user: false
    trust_remote_user_port_control_header: X-NETDISCO_USER_TEMPLATE
    trust_remote_user_port_control_re: port_control
    trust_remote_user_admin_header: X-NETDISCO_USER_TEMPLATE
    trust_remote_user_admin_re: admin

# Apache
    #RequestHeader set X-NETDISCO_USER_TEMPLATE "port_control,admin"
    RequestHeader set X-NETDISCO_USER_TEMPLATE "port_control"

This will then set the properties of the dynamically created user, depending on what was in the header. Only supports all-or-nothing port_control or admin flags, no portctl_roles

rc9000:
Shelfving this until #1353 has materialized

PR #1321, opened 2025-04-21
Add branding text option to web UI

bmtwl:
Include an optional branding configuration option.
Adding a branding_text option to the deployment.yaml will allow customization of the header.
eg: `branding_text: Netdisco XYZ Corporate Office`
if this option is not set, it will default to "Netdisco"

### DIFF
--- a/share/config.yml
+++ b/share/config.yml
@@ -25,6 +25,7 @@ tenant_databases: []
 # ------------
 
 domain_suffix: []
+branding_text: Netdisco
 no_auth: false
 suggest_guest: false
 navbar_autocomplete: true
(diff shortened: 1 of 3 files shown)


ollyg:
Hi @bmtwl thanks for the patch! good idea. Couple of things I request to approach differently before merging:

- If you add a default for `branding_text` to `settings.yml` then no need for the IF/THEN/ELSE (suggest alongside the `login_logo`setting in the file) - much simpler just change "Netdisco" for the setting :)
- Why is the link changed? (`uri_for` bit) The change possibly breaks features like `tenant_databases`, and URI encoding.
- Could also change the window title but that's a little more work as it changes dynamically based on the page viewed - I'm happy to look into that if it'd be useful.

thanks again

bmtwl:
OK I've made the requested changes.
Hopefully it matches your coding style standards

ollyg:
thanks @bmtwl I'll merge now!

PR #1324, opened 2025-04-23
function to retrieve subnets from NXOS devices via cli

### LINKED ISSUE #1323: Subnets not retrieved on NXOS devices with VRFs
Hi,
we noticed that on our NXOS devices (N77c7706) only a couple of subnets are fetched during arpnip via SNMP,
probably because we use multiple VRFs.
 (We had the same issue with arpnip/macsuck via SNMP and resolved it by configuring SSH for those devices.)
IP and MAC addresses are gathered fine now using the SSHCollector module (NXOS.pm)
So I added a function there to retrieve the subnets via CLI as well.

I am happy to create a PR if you would consider merging it.

j-hoff:
This adds a function to the NXOS SSHCollector module which retrieves subnets
by getting the routing table for all vrfs via cli and then filters for attached nets and ignoring /32.

closes #1323

### DIFF
--- a/lib/App/Netdisco/SSHCollector/Platform/NXOS.pm
+++ b/lib/App/Netdisco/SSHCollector/Platform/NXOS.pm
@@ -166,6 +166,58 @@ EOF
     return $macentries;
 }
 
+sub subnets {
+    my ($self, $hostlabel, $ssh, $args) = @_;
+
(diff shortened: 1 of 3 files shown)


ollyg:
Hi @j-hoff thank you so much! this is great, will merge now for next release.

Issue #1348, opened 2025-06-11
Negation of ACL for prop:val ACL does not yield the correct result


jijybarnes:
Hello there,
I've written some ACLs to hide ports that are not standard physical ports for Cisco Switches :
```yaml
hide_deviceports:
  vendor:cisco:
  - op:and
  - '!port:TwentyFiveGigE\d/\d/\d+'
  - '!port:GigabitEthernet\d/\d/\d+'
  - '!port:FastEthernet\d/\d+'
  - '!port:TenGigabitEthernet\d/\d/\d+'
```
I'm using the negation operator, and I noticed that there isn't any port displayed in the Ports tab with those ACL (I tested on a Catalyst 9300L which has GigabitEthernetX/X/X and TenGigabitEthernetX/X/X ports).
After some digging and debugging in App::Netdisco::Util::Permission, the problem seems to be in this loop from the `check_acl` subroutine:
```perl
          # line 287 to 288
          my $prop  = $1;
          my $match = $2 || '';
          ... # skipping to line 317
          # line 317 to 335
          ITEM: foreach my $item (@$things) {
              last ITEM if $found;
              if (blessed $item) {
                  if ($neg xor ($match eq q{} and ! $item->can($prop))) {
                    return true if not $all;
                    $found = true;
                    last ITEM;
                  }
              }
              elsif (ref {} eq ref $item) {
                  # empty or missing property
                  if ($neg xor ($match eq q{} and ! exists $item->{$prop})) {
                    return true if not $all;
                    $found = true;
                    last ITEM;
                  }
              }
            }
```
To be precise, it is this line that makes every negated prop:val rule check switch `$found` to `true` :

```perl
if ($neg xor ($match eq q{} and ! $item->can($prop)))
```

From what I understand, this condition is here in case an item is missing the property that the rule wants to check, and the condition matches an empty string. 
However, if the property exists and the condition is not empty, negating the rule will make the item match it.
For instance, the string `GigabitEthernet1/0/9` should not be matched since it matches '!port:GigabitEthernet\d/\d/\d+' which is negated. 

In this example,
- $neg equals to `1`
- the expression `$match eq q{}` equals to `false`
- the expression `! $item->can($prop)` will also be equal to false since $item->{$prop} exists here ($prop equals "port")

The entire expression is ` 1 ^ (0 & 0)` which yields 1, making it match a rule that shouldn't be matched.

I think the solution would be to remove the $neg variable from the expression, as it allows this following statement to be true :

[...]

PR #1353, opened 2025-06-17
Rework of portctl and roles

jijybarnes:
Hello ! 
This is a rework of how porctl is implemented I've been working on for work. We plan on using Netdisco to delegate permission to administrate switches to multiple IT teams.

This rework gets rid of the use of deployment.yml to declare roles and associated ACLs and moves it to the DB, to allow a more dynamic gestion (and to avoid restarting Netdisco each time we change portctl_role in deployment.yml)

Here's a quick summary of the changes : 

Added 3 new tables to move portctl roles and ACLs to the database in order to avoid the heavy server restart when updating the deployment.yml file. 
- PortctlRole : list of the roles created on netdisco
- PortctlRoleDevice: associates which device a role can admin
- PortctlRoleDevicePort: stores which ports a role is allowed to admin on a device they can admin (but mostly those the role can't admin, to save some DB space)

Added 2 new AdminTask pages ;
- AdminTask::RolePerm : administration of  roles and which devices they can admin
- AdminTask::DevicePortctl : administration of which ports a role is allowed to admin on the devices they can admin

Reworked  Util::Web accordingly:
- Added query string `q` variable for admin_tasks
- Modified hook->{user_has_role} to point to the database
Reworked Util::Port accordingly:
- Modified port_acl_by_role_check subroutine to use the database

Edit : I've also added the change to Permission.pm (see issue #1348), tests on preprod server seems to be coherent on my end.

Have a good day !


jijybarnes:
Hello again,
Don't worry, I'm not interpreting these notes negatively. On the contrary, I am quite happy to have things to fix.

I'm currently working on a commit that could answer most of your remarks. Here's my direct answer to your notes:

- **rename the schema update App-Netdisco-DB-Portctl-Update.sql to use a number sequence**
  - Agreed, that's an oversight on my part: I forgot to rename the file.
- **revert README.md change  + revert Util/Permission change**
  - I did both reverts in my last commit, Util/Permission and README.md are  the same as the one on the main branch.
- **default roles should exist in portctl_role table?**
  - The "_global_" role name exists only in Users.pm. When assigning this role to a user, an empty string is stored as role and the port_control attribute is set to true (cf. AdminTask/Users.pm:56).
- **PortctlRoleDevicePort can_admin text instead of boolean (it is in the underlying sql)**
   - Another oversight, will fix this.
- **why does ResultSet::PortctlRoleDevice role_can_admin not filter for admin**
  - It's actually because I thought that the can_admin field of the table isn't really relevant. Only devices that can be administered by X role should be in this table (since the default policy is set to deny admin permissions to a role). Please, tell me how you feel that this attribute would still be relevant.
- **ResultSet::PortctlRoleDevicePort can be deleted**
  - I completely agree.
- **Util/Port.pm is OK but cleanup perl and add back the deleted comments** 
  - Fixed the logic of my port_acl_by_role_check to check first if the user can_admin the device, then check if the user can administrate the port. 
- **AdminTask::DevicePortctl**
  - **why has duplicate port_acl_by_role_check** 
     - It was originally here because the one in Util/Port.pm didn't allow for a simple role permission check (not checking the current user portctl/admin permissions). It can and will be removed, a simple add of a parameter to specify if we are passing a role name or a user to the function signature in Util/Port.pm should do the trick.
  - **deviceportctl does not safeguard for missing q param (param safety generally needs reviewing)**
    - Error code 400 will be returned if param q isn't supplied. Otherwise, if q isn't an existing role name, an empty set will be returned
  - **deviceportctl doesn't need to search_for_device, can get_device instead.** 
    - Agreed 100%.
  - **needs transaction and error checking on the DB stuff** 
    - Wrapping DBIC operations in txn_do should do it, like in AdminTask/Users.pm 
  - **seems quite expensive and not scalable to load all the devices each time (try typeahead instead?)**
    - It is expensive. I haven't played with TypeAhead yet, I will look into it.
- **need to check license/owner on the images**
  - I couldn't find enough information on the license as it is present on multiple websites under multiple licenses. To be safe, I'll just replace it with something that is clearly free-to-use.
Example : https://www.svgrepo.com/svg/456211/ethernet-port. It is under MIT License (https://www.svgrepo.com/page/licensing/#MIT).

- **check impact of .nd_collapse-vlans, .nd_collapse-devices (where else is collapse-devices used?)**
  - .nd_collapse-vlans is used in Device/Ports.pm for the v_member column. As .nd_collapse-devices is a direct copy of the aforementioned class, it shouldn't pose any issue.

- **magic role name global can be config item or some other way**
  - Since '_global_' role name only exists in Users.pm and is equivalent to having no role set and the port_control attribute set to true, I don't think it needs to be stored somewhere.

- **idea: keep existing config support with Configuration.pm merging?? so people can use ACLs/prefixes**
A config directive and a little change to Util/Port.pm checking for this value would fulfill this need: 
```perl
sub port_acl_by_role_check {
  my ($port, $device, $user) = @_;

  my $mode = setting('permission_mode'); # new config directive (default would be 'config': ACLs defined in deployment.yml)
  if ($mode eq 'hybrid') {
    return (database_port_acl_by_role_check($port, $device, $user) or
            config_port_acl_by_role_check($port, $device, $user));
  }
  elsif ($mode eq 'database') {
    return database_port_acl_by_role_check($port, $device, $user);
  } # use ACLs defined in DB
  else {
    return config_port_acl_by_role_check($port, $device, $user);
  } # use ACLs defined in deployment.yml
}

```
However, config_port_acl_by_role_check must return false if $role is defined and the config doesn't have an entry for the role.
It shouldn't change the function's behavior if the user has all portctl permissions since $role would be an empty string.
```perl
    ...
    elsif ($role) {
        # the config does not have an entry for user's role
        return false; # that line is currently : return true if user->port_control;
    }
    ...
```


Anyways, thanks a lot for your initial feedback! I'll continue working on these changes and test them locally.
Cheers!

ollyg:
Hi @jijybarnes are you able to un-protect the master branch on your fork, so that I can push changes?

ollyg:
copied to #1402 to support better shared development, and closed

PR #1402, opened 2025-09-28
portctl and roles in the database (was #1353)

ollyg:


jijybarnes:
Hi @ollyg,
That's fine by me.
I'll use this branch.

jijybarnes:
Hey @ollyg,
I've built Netdisco from this branch and tested some stuff. 
There's some weird behavior with host_groups containing hostnames and the check_acl function.
Once the ACLs contained by the group are evaluated by the check_acl function, they aren't matched by anything and end up triggering this following snippet of the function around line 406 :
```perl
  RULE: foreach (@$config) {
     ...
      next RULE unless $addr; 
      # this is matched sincce
      # my $addr = NetAddr::IP::Lite->new($real_ip); on line 168
      # is the IP address of the device the user is currently trying to see
     # example, @$config is ['test-device-sw.example.com']
      my $ip = NetAddr::IP::Lite->new($rule)
        or next RULE; # not ip addresses, next rule
    ...
}
```
The same behavior can be observed with both ACLs from the database and from the config file.
This problem doesn't not occur with IP addresses, ranges, groups. It just occurs with standalone hostnames and regexes.

I'll investigate it further tomorrow and think of something.

ollyg:
thanks @jijybarnes that reminds me I have one more item to test with running jobs at cli with acls, I will do that too, so no rush

jijybarnes:
Hi @ollyg,
The ACL processing does PTR record search on potential the IP address of the device but I never see this code being called in this case. After some debugging, I've noticed that `ref $qref` evaluates to be a Regexp reference but `ref $rule` is an empty string (since rule is not a reference but a plain string in this case) :
```perl
   ...
  my $qref = ref qr//;

  RULE: foreach (@$config) {
      my $rule = $_; # must copy so that we can modify safely
      next RULE if !defined $rule or $rule eq 'op:and';

      if ($qref eq ref $rule) { 
          # if no IP addr, cannot match its dns
          next RULE unless $addr;

          $name = ($name || hostname_from_ip($addr->addr, $ropt) || '!!none!!');
      ...
```

Maybe to avoid any surprise, a better implementation of my patch would be to check for an existing DNS entry in the DB and if if doesn't exist, perform this DNS query and then compare the result to the rule. I'd need to change the condition of the aforementioned if statement too (i'll think of something). Let me know what you think about that.

Lastly, concerning this alternative ACL builder, if I understood correctly, it would store the IP address of the device instead of its FQDN for the device_acl (but would still displays the name of the device in the page). In my opinion, this is a good alternative. It would guarantee that a device is matched by its IP address while using names and regex expressions remain supported. I can think of some cases where that would be really useful (devices discovered by their Loopback address and interconnections addresses).

ollyg:
Hi @jijybarnes I've implemented the alternative IP/FQDN handling, and pushed to this branch

Issue #1396, opened 2025-09-20
Arista EOS Arpnip ssh collector module


robertlynch3:
I don't know if this is an issue for Netdisco or [SNMP-Info](https://github.com/netdisco/snmp-info/tree/master).
When running an arpnip on an Arista EOS device, we are not getting any ARPs for IP in VRFs other than the default. 

``` 
[netdisco@netdisco ~]$ bin/netdisco-do arpnip -d leaf5.mydomain.com -D
[4084954] 2025-09-20 13:13:40  info App::Netdisco version 2.089002 loaded.
[4084954] 2025-09-20 13:13:40  info arpnip: [192.168.25.20] started at Sat Sep 20 09:13:40 2025
[4084954] 2025-09-20 13:13:41 debug arpnip: running with timeout 600s
[4084954] 2025-09-20 13:13:41 debug //// CHECK \\\\ phase
[4084954] 2025-09-20 13:13:41 debug ⮕ worker Internal::BackendFQDN p1000000
[4084954] 2025-09-20 13:13:41 debug ⮕ worker Internal::SNMPFastDiscover p1000000
[4084954] 2025-09-20 13:13:41 debug running with configured SNMP timeouts
[4084954] 2025-09-20 13:13:41 debug ⮕ worker Arpnip p0
[4084954] 2025-09-20 13:13:41 debug ⬅ (done) arpnip is able to run
[4084954] 2025-09-20 13:13:41 debug //// EARLY \\\\ phase
[4084954] 2025-09-20 13:13:41 debug ⮕ worker Arpnip::Nodes p0 "prepare common data"
[4084954] 2025-09-20 13:13:41 debug //// MAIN \\\\ phase
[4084954] 2025-09-20 13:13:41 debug ⮕ worker Arpnip::Nodes p1000000
[4084954] 2025-09-20 13:13:41 debug ⬅ (info) skip: arp table data supplied by other source
[4084954] 2025-09-20 13:13:41 debug ⮕ worker Arpnip::Nodes p200
[4084954] 2025-09-20 13:13:41 debug ⬅ (info) skip: driver or action not applicable
[4084954] 2025-09-20 13:13:41 debug ⮕ worker Arpnip::Nodes p100
[4084954] 2025-09-20 13:13:41 debug snmp reader cache warm: [192.168.25.20]
[...]

PR #1404, opened 2025-09-29
Arista EOS SSHCollector

robertlynch3:
Created an SSHCollector for Arista EOS to work around #1396.


ollyg:
looks great, many thanks!

PR #1493, opened 2026-04-11
Fix/hook and jobqueue improvements

Bierchermuesli:
stumbled on that while I had a wrong config syntax: 

  - Validate type field in hook config before queuing: missing or unknown types now log a clear warning and skip the hook instead of producing a Perl uninitialised value warning and a cryptic _"failed to report from any worker"_ error in the UI

and think this is a small ui improovment:
  -  Job queue table: Tune "Device" column to "Device / Info" and show the first chars of the log (with full text on hover) for jobs without a device, giving visibility into hook and other non-device job results without needing to expand the details row:
image

### DIFF
--- a/lib/App/Netdisco/Util/Worker.pm
+++ b/lib/App/Netdisco/Util/Worker.pm
@@ -14,6 +14,17 @@ our @EXPORT = ('queue_hook');
 
 sub queue_hook {
   my ($hook, $conf) = @_;
+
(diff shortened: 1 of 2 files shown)


ollyg:
Looks good to me, many thanks!
Just a small note: in general I prefer each change or feature to be in its own patch (this is two in one) but it's not a big deal; I appreciate greatly the contribution!

PR #1499, opened 2026-04-20
Feature/ports Tooltips and Brush up

Bierchermuesli:
hey again

I hope it is ok to have a little brush up on the Device cell: 

   - Add hover tooltips to all icons (for dumb new users like us)
   -  slick one line per device/entry
   -  combine CDP/MAC entriy for APs/Phone into one line, remove duplicated information
   - Switches `hostname@port` format (IMHO remote port is irrelevant for APs)
   - Show `remote_type` for WAPs and phones
     -   _Edit: noticed afterwards, thi scan be enabled with "Remote Advertisement". Drop?_
   - Add `icon-laptop` to generic client to ensure a common listing style
   - remove odd `<br/>&nbsp;` 
   - Fix `node.wireless_port` check (was `node.wireless.defined` — has_many ResultSet is always truthy)

Feel free to tear apart or adjust. 

Before
list

After
after

PR #1505, opened 2026-05-04
From #1499 - node display improvements

ollyg:


ollyg:
Hi, this was merged into the test branch in netdisco repo. I made a couple of fixes and changes, and then merged to master branch. I did revert the "@" as I think hyphen looks nicer ;-) but the rest stayed, plus added a sidebar option to get DNS+IP or just DNS as you had in the patch

Issue #1422, opened 2025-10-21
Discover/Arpnip/Macsuck try multiple community strings


ralv3967:
The audit logs of some of our devices show a large number of failed SNMP connections, all from Netdisco. We use multiple community strings, all of which are stored in deployment.yaml. If I delete a device and then add it again, all CSs are still tried during subsequent actions.

ollyg:
Hi @ralv3967 so is it just for this device you have the problem, or a set of devices, or all of them? You say about half your devices have entries in the community table ... do they have the same problem?

ralv3967:
I compared the syslog entries with the entries in the community table. There are no entries in the table for the devices reporting failed SNMP logins. However, since Discover, etc., works for all of the 1,000-plus devices, I would expect to find an entry in the community table for all of them as well.

ralv3967:
Okay, I need to correct that a bit. I've now looked through the syslogs for the last 30 days and compared all the suspicious devices there with the community table. There are a few devices that have a correct entry in both the snmp_comm field and the community table, but still report failed logins.

ollyg:
Hi @ralv3967 I'm looking at this now, thanks for your patience. One thing I forgot to mention before is that adding access control lists might well help, if you can say which community string is to be used by which devices. That is, to use the `only` and `no` options in each `device_auth` stanza to limit each to the relevant devices.

ollyg:
OK first thing is that the `community` table will never be used, because there's no `tag` field set on `device_auth`:

```
            community   "COM1",
            driver      "snmp",
            no          [],
            only        [
                [0] "group:__ANY__"
            ],
            read        true,
            tag         "",
            write       true
```

So now we are down to why the `snmp_comm` field is not used when it has a value.

ollyg:
Hi @ralv3967 I found the cause, which was a regression as part of 9e5f3337b. Fix will be in the next release.

(I believe that well-intentioned change allowed users to more easily change the v2 community string or change from v2 to v3, but I would argue `device_auth` tags are a better way to achieve that.)

PR #1549, opened 2026-06-11
Add per-job param to discover via subaction

Bierchermuesli:
In environments with many mixed SNMP v3 credential profiles, the fast-pass (200ms per credential) fails for the correct credential because the device gets overwhelmed by rapid auth attempts before it is reached. Also related to #1422.

It also provides a non-destructive fix for the "stuck on wrong community string" scenario described in the Troubleshooting wiki: instead of deleting and re-discovering a device, you can force a discover with the correct tag to update the cached credential.

#### Supported `extra` params for discover

A plain string is treated as `snmp_tag` shorthand. JSON allows any combination of the following:

| param | type | description |
|---|---|---|
| `snmp_tag` | string | Use only the `device_auth` entry with this tag; skip all others and bypass cached-tag fast-probe. Must match a tag defined in `device_auth` config. |
| `snmptimeout` | integer (µs) | Override global `snmptimeout` for this job. Useful for high-latency or satellite-linked devices. |
| `snmpretries` | integer | Override global `snmpretries` for this job. |
| `bulkwalk_repeaters` | integer | Override global `bulkwalk_repeaters` for this job. Useful to reduce UDP fragmentation on high-latency links. |
| `skip_neighbor_queue` | bool | Store neighbor topology but do not queue new discovers for unknown neighbors. Prevents recursive discover cascades when manually triggering a one-off discover. |

## Usage

```bash
# plain string shorthand (snmp_tag only)
netdisco-do discover -d 192.0.2.1 -e site-a-creds

# explicit JSON with multiple params
netdisco-do discover -d 192.0.2.1 -e '{"snmp_tag": "site-a-creds"}'
netdisco-do discover -d 192.0.2.1 -e '{"snmptimeout": 3000000, "bulkwalk_repeaters": 5}'
netdisco-do discover -d 192.0.2.1 -e '{"snmp_tag": "site-a", "skip_neighbor_queue": true}'
```

```python
requests.post('/api/v1/queue/jobs',
  json=[{"action": "discover", "device": "192.0.2.1", "extra": "site-a-creds"}])

requests.post('/api/v1/queue/jobs',
  json=[{"action": "discover", "device": "192.0.2.1",
         "extra": '{"snmptimeout": 3000000, "skip_neighbor_queue": true}'}])
```

## Tests / Todo

- [x] Discover a device with `-e <tag>` and confirm only that credential is tried in debug log
- [x] Discover with JSON `-e '{"snmp_tag": "..."}'` and confirm same
- [x] Discover without `-e` and confirm normal credential cycling behaviour is unchanged
- [x] Discover with an unknown tag and confirm fallback to normal behaviour
- [ ] Confirm "stuck on wrong community" scenario resolves without deleting the device
- [x] Discover a high-latency device via queued job and confirm slow connect fallback runs
- [x] Discover with `skip_neighbor_queue: true` and confirm neighbors are stored but none queued
- [ ]  Wiki Update

### DIFF
--- a/lib/App/Netdisco/Backend/Job.pm
+++ b/lib/App/Netdisco/Backend/Job.pm
@@ -1,6 +1,6 @@
 package App::Netdisco::Backend::Job;
 
-use Dancer qw/:moose :syntax !error/;
+use Dancer qw/:moose :syntax !error !params/;
 use aliased 'App::Netdisco::Worker::Status';
 
 use Moo;
--- a/lib/App/Netdisco/Backend/Job.pm
+++ b/lib/App/Netdisco/Backend/Job.pm
@@ -4,6 +4,7 @@ use Dancer qw/:moose :syntax !error/;
 use aliased 'App::Netdisco::Worker::Status';
 
 use Moo;
+use Try::Tiny;
 use Term::ANSIColor qw(:constants :constants256);
 use namespace::clean;
 
@@ -226,5 +227,18 @@ priority of the job and is used by L<MCE> when managing its job queue.
 =cut
 
 sub extra { (shift)->subaction }
+
+=head2 params
+
+Parses the C<subaction> field as JSON and returns a hashref of parameters.
+Returns an empty hashref if subaction is empty or not valid JSON.
+
+=cut
+
+sub params {
+  my $job = shift;
+  return {} unless $job->subaction;
+  return try { from_json($job->subaction) } catch { {snmp_tag => $job->subaction} };
+}
 
 true;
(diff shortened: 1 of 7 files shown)


ollyg:
it's good! I will merge. thank you! just two thoughts, for future enhancement:

- I'd make the JSON input a wide-open interface for config, what I mean is, take the JSON converted to dict, and MERGE with existing loaded config (which is itself a merge of Netdisco defaults and user `deployment.yml`), can use the built in Dancer merger for that. Then the `snmptimeout`, `snmpretries`, `bulkwalk_repeaters` need no special handling in `SNMPFastDiscover.pm`.
- I've a feeling (but not checked) that `skip_neighbor_queue` can be achieved using existing config, but that config requires the device IP in an ACL, which is annoying for new discoveries.  Wonder if a new setting `set_for_this_device` could be added to contain a full config settings tree which is also merged and overrides `deployment.yml`.

ollyg:
just to note that in the implementation released, changed `snmp_tag` to be `device_auth_tag_hint`

PR #1550, opened 2026-06-11
add permanent API tokens and (dummy) Bearer auth support

Bierchermuesli:
In environments where service accounts need stable, non-expiring API credentials, session tokens (which expire after `api_token_lifetime`) are not suitable. This adds permanent tokens with optional IP restrictions.

- Adds `token_no_expire` boolean and `token_allowed_ips text[]` columns to `users` table (DB migration 96→97)
- Adds `token_auth_only` boolean column — a semantic marker for accounts intended for token-only access (no password). Enforcement is already provided by the null password, but the flag drives the GUI display and is exposed via `GET /api/v1/users` so API consumers can identify service accounts without inspecting password state. Note: `token_auth_only=true` implies `token_no_expire=true` — a passwordless account with a short-lived token would be permanently locked out once it expires. The GUI enforces this by only offering a single "Permanent Token" option.
- Permanent tokens bypass the `api_token_lifetime` expiry check in both `validate_api_token` and the `UserRole` SQL view
- `Authorization: Bearer <token>` is now accepted alongside the existing raw `Authorization: <token>` format — no OAuth machinery, just the standard HTTP scheme. Both styles continue to work.
- Per-user CIDR IP allowlist: if `token_allowed_ips` is set, API token requests from outside the listed prefixes are rejected (applies to token validation only, not web GUI login)
- `getapikey` gains `-p permanent` to create a non-expiring token and `-p revoke` to clear all tokens for a user
- `/login` API endpoint now accepts a JSON body `{"permanent": true, "allowed_ips": ["10.0.0.0/8"]}` and returns `{api_key, permanent, allowed_ips}` — gated by `allow_permanent_tokens: false` config flag
- `POST /api/v1/user` — idempotent service account provisioning: creates user if absent (null password, token-auth only), issues or rotates token, sets permanent flag and IP allowlist
- `GET /api/v1/users` — list all users with roles and token status (hint, permanent flag, allowed IPs)
- GUI service account management: auth method dropdown has a single **Permanent Token** option (sets `token_auth_only` + `token_no_expire`), dedicated token-generate button, allowed IPs field

image

## Usage

```bash
# Via netdisco-do (headless/CLI)
~/bin/netdisco-do getapikey -e grafana-user -p permanent

# Via /login API (swagger UI or curl)
curl -X POST /login -u user:pass \
  -H 'Accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{"permanent": true, "allowed_ips": ["192.168.1.0/24"]}'

# Via REST API (service account provisioning)
curl -X POST /api/v1/user \
  -H 'Authorization: Bearer <admin-token>' \
  -H 'Content-Type: application/json' \
  -d '{"username": "grafana-svc", "permanent": true, "allowed_ips": ["10.0.0.0/8"]}'

# Use the token — both formats accepted
curl /api/v1/... -H 'Authorization: Bearer <token>'
curl /api/v1/... -H 'Authorization: <token>'
```

## Config

```yaml
allow_permanent_tokens: false  # set true to allow permanent tokens
```

## Test plan

- [x] Normal session token still expires after `api_token_lifetime`
- [x] `token_no_expire: true` bypasses expiry in both `validate_api_token` and UserRole SQL
- [x] `Authorization: Bearer <token>` accepted; raw `Authorization: <token>` without prefix also still works
- [x] `token_allowed_ips` rejects requests from outside the listed CIDRs
- [x] Web GUI login not affected by `token_allowed_ips`
- [x] `getapikey -p permanent` sets `token_no_expire = true`; `-p revoke` clears all token fields
- [x] `/login` with `permanent: true` only works when `allow_permanent_tokens: true` in config
- [x] `POST /api/v1/user` creates account and issues token idempotently
- [x] `GET /api/v1/users` returns user list with token hints and flags
- [x] GUI: Permanent Token dropdown sets token_auth_only + token_no_expire; token button always issues permanent token
- [ ] Wiki update

### DIFF
--- a/bin/netdisco-do
+++ b/bin/netdisco-do
@@ -518,11 +518,21 @@ L<App::Netdisco::Worker::Plugin::MakeRancidConf> for configuration needs.
 
 =head2 getapikey
 
-Generates an API key for the supplied username. See the 
-L<API doc|https://github.com/netdisco/netdisco/wiki/API> for further 
+Generates an API key for the supplied username. See the
+L<API doc|https://github.com/netdisco/netdisco/wiki/API> for further
 information.
 
- ~/bin/netdisco-do getapikey -e the_username 
+ ~/bin/netdisco-do getapikey -e the_username
+
+To create a permanent token (no expiry) pass C<permanent> in the C<-p> parameter.
+The token can be used in the C<Authorization: Bearer E<lt>tokenE<gt>> HTTP header
+and is accepted by any tool supporting standard Bearer authentication (Grafana, curl, etc.).
+
+ ~/bin/netdisco-do getapikey -e the_username -p permanent
+
+To revoke all API tokens for a user:
+
+ ~/bin/netdisco-do getapikey -e the_username -p revoke
 
 =head2 dumpconfig
 
--- a/lib/App/Netdisco/DB.pm
+++ b/lib/App/Netdisco/DB.pm
@@ -11,7 +11,7 @@ __PACKAGE__->load_namespaces(
 );
 
 our # try to hide from kwalitee
-  $VERSION = 96; # schema version used for upgrades, keep as integer
+  $VERSION = 97; # schema version used for upgrades, keep as integer
 
 use Path::Class;
 use File::ShareDir 'dist_dir';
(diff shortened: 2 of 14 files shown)


ollyg:
looks great! I will merge. Many thanks!! for future enhancements:

- could use `check_acl` on `token_allowed_ips` (supports more formats more safely, and keeps list checking in one place)
- could use the web-based ACL manager feature if I write it, to manage `token_allowed_ips`

Also noticed that if token is revoked, then `token_allowed_ips` is emptied. Be sure to document this behaviour.

PR #1565, opened 2026-06-19
option to split scheduler and worker for HPA

Bierchermuesli:
## Summary

Adds two environment variables to `netdisco-backend-fg` to enable splitting the backend into separate roles — a single scheduler replica and a horizontally scalable worker replica set.

- **`NETDISCO_NO_SCHEDULER=1`**: disables the scheduler task entirely, making the pod a pure worker node safe to scale with HPA without causing duplicate job scheduling
- **`NETDISCO_WORKERS_TASKS=N`**: overrides the `workers.tasks` config value at runtime, allowing per-Deployment worker tuning without a separate ConfigMap

This is reflected in [netdisco-helm chart **0.4.0**](https://github.com/netdisco/helm-charts/pull/6)  which introduces separate `scheduler` and `worker` Deployments.

## Example k8s setup

```yaml
# scheduler Deployment — 1 replica, no special env
# handles all cron/schedule jobs, never scaled

# worker Deployment — N replicas, HPA-scalable
env:
  - name: NETDISCO_NO_SCHEDULER
    value: "1"
  - name: NETDISCO_WORKERS_TASKS
    value: "20"
```

Startup log is also more verbose depending on the roles:
 - `warn mode: worker-only | workers: 12 | host: netdisco-backend-....`
 - `warn mode: scheduler | host: netdisco-scheduler-... `
 - `warn mode: scheduler+worker | workers: 12 | host: ...` 
## Test plan

- [x] Start backend without env vars — verify `scheduler+worker` mode, schedule jobs fire normally
- [x] Start backend with `NETDISCO_NO_SCHEDULER=1` — verify `worker-only` mode, no scheduler task spawned
- [x] Start backend with `NETDISCO_WORKERS_TASKS=5` — verify worker count overridden
- [x] Run two worker-only pods simultaneously — verify no duplicate job scheduling

### DIFF
--- a/bin/netdisco-backend-fg
+++ b/bin/netdisco-backend-fg
@@ -69,11 +69,11 @@ if (setting('workers')->{'no_scheduler'}) {
   shift @task_names; shift @task_workers; shift @task_roles;
 }
 
-warning sprintf "mode: %s | workers: %s | host: %s",
+warning sprintf "mode: %s%s | host: %s",
   (setting('workers')->{'no_scheduler'} ? 'worker-only'
    : $max_workers == 0                  ? 'scheduler'
    :                                      'scheduler+worker'),
-  $max_workers,
+  ($max_workers ? " | workers: $max_workers" : ''),
   (setting('workers')->{'BACKEND'} || 'unknown');
 
 mce_flow {
--- a/bin/netdisco-backend-fg
+++ b/bin/netdisco-backend-fg
@@ -70,7 +70,9 @@ if (setting('workers')->{'no_scheduler'}) {
 }
 
 warning sprintf "mode: %s | workers: %s | host: %s",
-  (setting('workers')->{'no_scheduler'} ? 'worker-only' : 'scheduler+worker'),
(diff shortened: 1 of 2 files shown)


ollyg:
this is fine.

BTW, setting `schedule: null` causes the Scheduler worker to quit with a log message (see [unless (setting('schedule')) {](https://github.com/netdisco/netdisco/blob/4f30a05157f89204258aabb3788a148cfd56749a/lib/App/Netdisco/Backend/Role/Scheduler.pm#L50)), same effect as `NETDISCO_NO_SCHEDULER`.

so, this could be implemented by adding NETDISCO_NO_SCHEDULER check in the begin and body parts of the scheduler, alongside undef, albeit with the overhead of starting then stopping the worker. equally, could move the undef check out to where NETDISCO_NO_SCHEDULER is. 

I don't mind leaving as-is, or changing, up to you @Bierchermuesli. many thanks!

Bierchermuesli:
You're right, haven't noticed that. let me know if we shall simplify it.

PR #1570, opened 2026-06-23
support NETDISCO_BACKEND_NAME env var to set shared backend identity

Bierchermuesli:
this one was missed for #1565:  In replicated deployments (e.g. Kubernetes HPA), each pod gets a unique FQDN which becomes the `workers.BACKEND` key in `DeviceSkip`. With `max_deferrals` set, each replica tracks skip counts independently, so a device is retried `N*max_deferrals` times instead of `max_deferrals` times total. So basically it reverse the idea of backends in a classic deployment :-)


ollyg:
many thanks!

PR #1582, opened 2026-07-06
Add discover/macsuck/arpnip_min_age settings to skip recently-walked devices

Bierchermuesli:
What you think about this?

- Add `discover_min_age`, `macsuck_min_age`, `arpnip_min_age` settings (minutes) that skip queuing a walk job for a device if it was already walked within that window

In event-driven deployments (e.g. jobs queued mid-day from API), the same device can be re-queued for discover/macsuck/arpnip far more often than needed. Without a minimum-age guard, every trigger re-walks the device even if it was just walked <duration> ago, wasting worker time and adding needless load on the device and the queue. These settings let devices that were recently walked skip re-queuing, so the queue only does work when it's actually stale.

Defaults to 0 (behaviour as today), however I would suggest these defaults (which would fit to the schedule default  we ship): 
`discover_min_age=1440` (24h), `macsuck_min_age=60`, `arpnip_min_age=60` 

## Test plan
- [x] Verified `is_discoverable_now` / `is_arpnipable_now` / `is_macsuckable_now` correctly compare `since_last_*` (seconds) against the minute-based setting
- [x] Confirmed default of `0` preserves existing behavior for anyone not setting these

### DIFF
--- a/lib/App/Netdisco/Configuration.pm
+++ b/lib/App/Netdisco/Configuration.pm
@@ -391,9 +391,9 @@ setting('workers')->{'max_deferrals'} ||= (2**30);
 setting('workers')->{'retry_after'}   ||= '100 years';
 
 # 0 seconds means no minimum age filter (all devices eligible for walk)
-config->{'discover_min_age'} ||= 86400;
-config->{'macsuck_min_age'}  ||= 3600;
-config->{'arpnip_min_age'}   ||= 3600;
+config->{'discover_min_age'} ||= 0;
+config->{'macsuck_min_age'}  ||= 0;
+config->{'arpnip_min_age'}   ||= 0;
 
 # schedule expire used to be called expiry
 setting('schedule')->{expire} ||= setting('schedule')->{expiry}
--- a/lib/App/Netdisco/Configuration.pm
+++ b/lib/App/Netdisco/Configuration.pm
@@ -391,9 +391,9 @@ setting('workers')->{'max_deferrals'} ||= (2**30);
 setting('workers')->{'retry_after'}   ||= '100 years';
 
 # 0 seconds means no minimum age filter (all devices eligible for walk)
-config->{'discover_min_age'} ||= '0 seconds';
-config->{'macsuck_min_age'}  ||= '0 seconds';
-config->{'arpnip_min_age'}   ||= '0 seconds';
+config->{'discover_min_age'} ||= 86400;
+config->{'macsuck_min_age'}  ||= 3600;
+config->{'arpnip_min_age'}   ||= 3600;
 
 # schedule expire used to be called expiry
 setting('schedule')->{expire} ||= setting('schedule')->{expiry}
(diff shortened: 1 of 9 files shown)


rc9000:
> Add discover_min_age, macsuck_min_age, arpnip_min_age settings (minutes) that skip queuing a walk job for a device if it was already walked within that window

it sounds good not to queue them unnecessarily but do we really need to change the unit on the existing setting? 

And maybe Worker/Plugin/XXXWalk could just call Device:is_xxx_now instead of doing its own calculations and spreading the decision logic around.

Just my hunch after some minutes of looking at it.

ollyg:
Hi, I agree with @rc9000 that the unit should not change, even though I agree with what I guess is the point that seconds is a bit useless from a user perspective. The design was lazy but I don't want a breaking change now.

I would also not change the defaults. For the vast majority of installations, there is little negative consequence of default zero, because the job is very quickly rejected after picking. You would have to have an awful lot of these to clog up the pollers.

For the other question from @rc9000 ...

The problem with running the results of WalkJobs through is_discoverable_now before queueing, is that is_discoverable_now will only check the config for the local host (the scheduler) which might not be the same config (ACLs etc) as the polling host. Not many people will have such a set-up but it is a possible set-up, and would mean the config on the scheduler has to be crafted more carefully so as not to be too liberal or iliberal (for that device discovery). Sorry, that was a bit long-winded 😅 .

Anyway, back to the PR, I don't see why not to put it into the SQL. So that part is fine, I think.

(p.s. in general I am not a fan of min_age settings, because if you set `discover_min_age=(1440*60)` and then discover a device mid-cycle, you actually wait another 1.5 cycles before it's scheduled again for discovery, which to me is worse than a poller picking and quickly rejecting a job, or running an extra poll.)

(edited for formatting fixes)

ollyg:
Oh, you asked about defaults, too. This is also an issue for the min_age because the jobs in a walk don't run at the same second, they're smeared out over some period of time. You can see this in the Poller Performance admin report - Time Elapsed column.

The min_age might well be less than the default even between two legitimate scheduled runs, because it is comparing job run to job run, not schedule to schedule.

Personally I would rather have occasional extra polling, than skipped polling, so I avoid the min_age settings altogether. YMMV of course and I welcome to learn about how others successfully use the feature, especially if the need is quite frequent (we could look at other ways of mitigating).

Bierchermuesli:
Fair points on both — happy to revert the unit back to seconds and keep the 0 defaults, not worth a breaking change over.

On the actual need: we're running an event-driven setup where jobs get queued from external triggers (webhooks/API) rather than a fixed schedule, so the same device can end up requeued for discover/macsuck/arpnip more often than a schedule would ever produce. Sometimes discovery runs multiple times a day (on device change in netbox). Without a floor, the poller just re-walks a device that was walked <n-times> ago, which adds up at higher volume than needed. 

If you'd still rather not carry this, happy to just drop it — not attached to it.

ollyg:
Hi @Bierchermuesli sounds good! if you change back to seconds I will review again, I'm sure we can merge this

ollyg:
@Bierchermuesli remove the change to Configuration.pm and we're good to merge

PR #1583, opened 2026-07-06
fix port lookup in API: uri_unescape and trim captured port name

Bierchermuesli:
Dancer's `captures()` does not URL-decode regex path segments, so the per-port API routes in `Web/API/Objects.pm` (`/api/v1/object/device/{ip}/port/{port}` and the nodes/power/etc sub-resources) were passing the raw, still-encoded port name straight to `->find()`.

Any port name containing a slash - basically every stacked/modular switch port like `Gi1/0/1`, sent URL-encoded as `Gi1%2F0%2F1` - never matched, and the API returned a 404 for a port that exists. SNMP-stored port names can also carry stray leading/trailing whitespace, which caused the same kind of false 404.

Example: `GET /api/v1/object/device/10.0.0.1/port/Gi1%2F0%2F1` returned 404 before this fix even though the port is in `device_port`. After decoding and trimming the captured value it resolves correctly.

The fix adds a small `_port_param()` helper that runs the captured port through `uri_unescape()` and strips whitespace before the DB lookup, and applies `uri_unescape()` to the `ip` capture in the same three spots for consistency.

### DIFF
--- a/lib/App/Netdisco/Web/API/Objects.pm
+++ b/lib/App/Netdisco/Web/API/Objects.pm
@@ -7,6 +7,9 @@ use Dancer::Plugin::Auth::Extensible;
 
 use App::Netdisco::JobQueue 'jq_insert';
 use Try::Tiny;
+use URI::Escape 'uri_unescape';
+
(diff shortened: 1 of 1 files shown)


ollyg:
good catch! thank you!

PR #1584, opened 2026-07-06
feat. make GET /api/v1/object/device batch ready with field selection

Bierchermuesli:
Adds `GET /api/v1/object/devices`, a bulk device list endpoint. Unlike `/api/v1/search/device`, which requires at least one filter param (`q`, `name`, `location`, etc.) and returns a fixed set of columns unless `seeallcolumns` is set, this endpoint returns the full device inventory (or a page of it) with a caller-chosen column list and no filtering required.

Main use case is comparing/syncing against external systems (CMDB, monitoring) where you want the whole inventory, or a specific subset of fields across all devices, rather than a filtered lookup. Supports `fields` (comma-separated column list, `all` for every column, plus a `device_auth_tag` join extra), `limit` and `offset` for paging large inventories.

Example: `GET /api/v1/object/devices?fields=ip,dns,vendor,device_auth_tag&limit=100&offset=200`.

Happy to fold this into `/search/device` instead if you'd prefer one endpoint - it just seemed like a bigger change since that route currently requires a filter and only supports the fixed/all column set, so relaxing that would affect every other search tab using the same dispatcher. Let me know if you'd rather go that route.

on the other side wondering if a simple q parameter like https://github.com/netdisco/netdisco/pull/1584/changes/2f85f24edc9efbe648234629166931f146d87efc could make sense or if we conflict with the idea... :)

### DIFF
--- a/lib/App/Netdisco/Transport/SNMP.pm
+++ b/lib/App/Netdisco/Transport/SNMP.pm
@@ -299,10 +299,11 @@ sub _try_connect {
 
   try {
       $snmp_args->{Offline} || debug
-        sprintf '[%s:%s] try_connect with v: %s, t: %s, r: %s, class: %s, comm: %s',
+        sprintf '[%s:%s] try_connect with v: %s, t: %s, r: %s, class: %s%s, comm: %s',
           $snmp_args->{DestHost}, $snmp_args->{RemotePort},
           $snmp_args->{Version}, ($snmp_args->{Timeout} / 1000000), $snmp_args->{Retries},
-          $class, $debug_comm;
+          $class,
+          ($comm->{tag} ? ', tag: '. $comm->{tag} : ''), $debug_comm;
       Module::Load::load $class;
 
       $info = $class->new(%$snmp_args, %comm_args) or return;
--- a/lib/App/Netdisco/Web/API/Objects.pm
+++ b/lib/App/Netdisco/Web/API/Objects.pm
@@ -15,11 +15,6 @@ swagger_path {
   path => (setting('api_base') || '').'/object/devices',
   description => 'Returns list of all known devices',
   parameters => [
-    q => {
-      in => 'query',
-      description => 'Filter by IP, dns, or name (substring match)',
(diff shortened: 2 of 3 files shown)


ollyg:
Hi @Bierchermuesli many thanks for the idea! I'm inclined to fold this into the existing `/api/v1/search/device` and here's why:

In fact the existing endpoint can return all results, if the user provides "`%`" as the "`q`" parameter (this is to the API... in the HTML frontend it changes "`*`" to "`%`" I think, to help the user).

I don't think it will be too intrusive to add `fields`, `limit`, and `offset` to the existing method. They would have sensible defaults which allow the existing HTML frontend to work as-is.

We could make "`q`" param optional on the condition that if it is not provided (all results) then `limit` and `offset` _must_ be provided. What do you think?

Also, for the `device_auth_tag` magic helper field, we can make an SQL JOIN to the community table but not prefetch, and then providing `community.snmp_auth_tag_read` will return the field in the JSON. Or we can keep the magic helper field (or both).

If you are OK with this, I don't mind implementing it.

Bierchermuesli:
> We could make "q" param optional on the condition that if it is not provided (all results) then limit and offset must be provided. What do you think?

This is perfect solution - i will look into that

Bierchermuesli:
Went ahead and fold that into `/api/v1/search/device` - `q` is now optional, but `limit`+`offset` are both required if you don't pass `q` or any filter param (so it doesn't accidentally full-table-scan). Added `fields`/`limit`/`offset`, carried over the `device_auth_tag` join. Dropped `GET /api/v1/object/devices`.

Tested it against our own instance (~10k devices) and it's working nice so far - or do you suggest a default limit?

@ollyg  are you OK with fading out `seeallcolumns` ? `fields` covers the same case plus arbitrary column lists, so it feels redundant to keep both long-term. I set  `seeallcolumns=true`  as alias to `fields=all`  for backwards compatiblity

ollyg:
@Bierchermuesli looks great! many thanks. I can merge and release.

PR #1591, opened 2026-07-13
new setting expire_nodeip_orphans

rc9000:
new setting expire_nodeip_orphans

 - set to true, expire immediately deletes node_ip entries without matching node entry - this is the original behavior and stays enabled by default
 - set to false, node_ip entries will remain in the table until a time-based limit from expire_nodes or expire_nodeip_freshness is met

This allows Netdisco to keep potentially interesting node_ip entries, even if the environment can not provide node entries


ollyg:
LGTM! many thanks

PR #1592, opened 2026-07-14
Allow any setting override at runtime from command line, api, or environment

ollyg:


ollyg:
I'd like to

- [x] make this work for both -e and -p param to netdisco-do/deployment.yml and not care which
- [x] make NETDISCO_WITH_CONFIGURATION take k=v,k=v as well as JSON
- [x] put the auth tag hint as the first tag to try rather than the only tag to try

PR #1607, opened 2026-08-05
Refuse to empty snmp_object when the MIB reports parse to nothing

jeneric:
## What happens today

`netdisco-do loadmibs` deletes every row of `snmp_object` and repopulates it
from `$mibhome/EXTRAS/reports/*_oids`. There is no check that the parse produced
anything, so when the reports yield zero objects the table is emptied and the
job exits 0.

Reproduced on a live install (PostgreSQL 18.4, App::Netdisco 2.101000):

```
loadmibs - loaded 0 objects from netdisco-mibs
loadmibs - removed 597248 oids
loadmibs - added 0 new oids
loadmibs: status done
```

The SNMP object browser is empty afterwards and nothing reports a failure.

## Why the reports parse to nothing

From netdisco-mibs 4.054 onward, `EXTRAS/reports/*` is under a git-lfs filter,
and GitHub tag archives do not resolve LFS. The files are present but contain
pointer stubs:

```
$ head -1 $mibhome/EXTRAS/reports/cisco_oids
version https://git-lfs.github.com/spec/v1
```

Every file in that directory matches the one filter, so this is all-or-nothing
rather than a partial-content case. Worth noting that 4.053 is not affected:
it ships no `*_oids` files at all, so `read_lines` croaks and the job already
fails loudly. Only 4.054 and later are silent.

## What this changes

Two commits.

**`loadmibs: read each report file once`** removes a double read. The `*_oids`
glob already returns `rfc_oids`, `net-snmp_oids` and `cisco_oids`, which are
then read a second time from the hardcoded list. `%seenoid` was already deduping
the parse, so the loaded rows are unchanged; this only stops re-reading the
files, of which `cisco_oids` alone is 27 MB.

**`loadmibs: refuse to empty snmp_object when nothing was parsed`** adds the
guard. When the parse yields no objects the delete and repopulate are skipped,
the reason is logged at error level, and the job's status becomes `error`, which
`bin/netdisco-do` turns into exit 1.

The two blocks after the guard still run. They act on `device_browser`, not
`snmp_object`: snapshot promotion reads the table the refusal just preserved,
and the legacy snapshot upgrade is unrelated work that should not become
collateral damage. That is why the guard calls `add_status` and lets the worker
continue, rather than returning `Status->error` immediately.

## Deliberately not fixed here

Two other things in this file turned up while investigating. Both are
independent of this bug, so both are filed as their own issues rather than
widening this PR. They are mentioned here because they are useful context while
reading the same function.

1. **#1610, `LoadMIBs.pm:27-31` filters full paths, not basenames.** Both `grep`s run
   before `map { (splitdir($_))[-1] }`, so `! m/\./` rejects any path containing
   a dot: give the mibhome a dotted path and `@maps` empties and every vendor
   report is silently skipped. A release tarball unpacked as
   `netdisco-mibs-4.054/` hits this. The `! m/^(?:EXTRAS)$/` filter on the line
   above is dead for the same reason, since a full path can never equal
   `EXTRAS`. The guard still fires correctly on a stubbed tree, because the
   three hardcoded reports are stubs too, but a mixed tree could load partial
   data and pass.
2. **#1611, `netdisco-do loadmibs -e <vendor>` replaces the whole table** with one
   vendor's objects, because the read is scoped to that vendor but the delete is
   not. The guard covers the empty case on that path but not the partial one.

## Root cause, for context

This is a guard, not a fix for the underlying situation. `snmp_object` looks
like it is mid-migration out of the MIB reports and into
`upstream-sources/bootstrap/netdisco-mibs-snmp-object.sql` in netdisco-mibs
(commit `57498439`), but netdisco was never wired to fetch it. Happy to work on
the consumer side of that separately if it would help.

## Testing

New file `xt/31-loadmibs-guard.t`, four subtests, runs with no database like the
rest of `xt/`:

- reports are LFS stubs: the job errors, says why, logs at error level, reads
  each report once, and issues no `delete` or `populate` against `snmp_object`
- reports parse to objects: the guard does not fire and the table is replaced
- a block after the guard dies: the refusal is still the job's outcome, and the
  later failure is recorded rather than swallowed
- the single-vendor path with a stub report: the guard covers it too

Verified by mutation rather than by inspection. Removing the guard, inverting
it, dropping either half of the refusal, returning early instead of continuing,
and reverting the dedupe each cause a failure.

Two things about the test that are worth knowing before reading it. It defines a
stand-in schema and installs it over the plugin's imported `schema`, a pattern
nothing else in `xt/` uses; it is there because this worker calls the database
unconditionally and the rest of `xt/` has no such dependency. And it overrides
the Dancer logger to `capture`, because the guard logs at error level and a
clean run would otherwise print the refusal every time; `DANCER_DEBUG` restores
the console.

`bin/netdisco-deploy` and the process exit code have no automated coverage. Both
were verified by hand.

## Operator impact

Anyone whose `loadmibs` currently succeeds against netdisco-mibs 4.054 or later
starts getting exit 1 instead of a silently emptied table. That is the point,
but it will stop a `set -e` cron or CI step that has been quietly destroying
data, so it is worth a release note.

The one in-tree caller, `bin/netdisco-deploy:288`, discards `system()`'s return
and prints `done.` regardless, so it does not yet notice. That is sent
separately, since it stands on its own: `loadmibs` can already fail for other
reasons and none of them reach the operator today. Either PR can go in without
the other.

## Two offers

- The guard uses `$job->add_status` so the blocks below still run.
  `add_status` does not otherwise appear in a worker plugin, only in framework
  code. If you would rather it read like the other ~30 `Status->error` sites, I
  can move those two `DeviceBrowser` blocks into a second
  `register_worker({ phase => 'store' })` in the same file and have the guard
  return `Status->error` directly. Behaviour and exit code are identical either
  way, so I left it alone rather than restructure on a guess.
- I have not added a `Changes` entry, since the recent ones look like they land
  in their own commits rather than inside a feature PR. Say the word and I will
  add one under `[BUG FIXES]`.

### DIFF
--- a/bin/netdisco-do
+++ b/bin/netdisco-do
@@ -549,6 +549,9 @@ Pass a device in C<-d> to display them:
 A requirement for web browsing of SNMP data (see C<snapshot> command), run
 this to load Netdisco MIBs into the database. It only needs to be done once.
 
+If the MIB reports parse to nothing, the job leaves the C<snmp_object> table
+unchanged, logs the reason, and exits non-zero.
+
 =head2 snapshot
 
 Builds and saves a data structure which Netdisco can use to mimic the device
--- a/lib/App/Netdisco/Worker/Plugin/LoadMIBs.pm
+++ b/lib/App/Netdisco/Worker/Plugin/LoadMIBs.pm
@@ -35,8 +35,12 @@ register_worker({ phase => 'main' }, sub {
       push @report, read_lines( catfile( $reports, "${vendor}_oids" ), 'latin-1' );
   }
   else {
+      my %seen_report = ();
+      my @to_read = grep { not $seen_report{$_}++ }
+                    (qw(rfc_oids net-snmp_oids cisco_oids), @maps);
+
       push @report, read_lines( catfile( $reports, $_ ), 'latin-1' )
-        for (qw(rfc_oids net-snmp_oids cisco_oids), @maps);
+        for @to_read;
   }
   
   my @browser = ();
--- a/lib/App/Netdisco/Worker/Plugin/LoadMIBs.pm
+++ b/lib/App/Netdisco/Worker/Plugin/LoadMIBs.pm
@@ -78,12 +78,24 @@ register_worker({ phase => 'main' }, sub {
   debug sprintf "loadmibs - loaded %d objects from netdisco-mibs",
     scalar @browser;
 
-  schema('netdisco')->txn_do(sub {
-    my $gone = schema('netdisco')->resultset('SNMPObject')->delete;
-    debug sprintf 'loadmibs - removed %d oids', $gone;
-    schema('netdisco')->resultset('SNMPObject')->populate(\@browser);
-    debug sprintf 'loadmibs - added %d new oids', scalar @browser;
-  });
+  if (not scalar @browser) {
+    my $refusal = sprintf
+      'loadmibs - refusing to empty snmp_object: read %d lines from %s but '
+    . 'parsed 0 objects', scalar @report, $reports;
+
+    # error() is what survives --quiet on the console. add_status() records the
+    # refusal now so the blocks below still run and cannot displace it.
+    error $refusal;
+    $job->add_status( Status->error($refusal) );
+  }
+  else {
+    schema('netdisco')->txn_do(sub {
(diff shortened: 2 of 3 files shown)


ollyg:
yes makes a lot of sense, thanks for the catch, as the oids files are not shipping right now. good tests!

PR #1615, opened 2026-08-09
Modernize the frontend dependencies: Bootstrap 5, Font Awesome 7, DataTables 2, jQuery 3

jeneric:
# Modernize the frontend dependencies: Bootstrap 5, Font Awesome 7, DataTables 2, jQuery 3

This replaces the frontend libraries under `share/public`. Bootstrap 2.3.1 dates
from 2013 and everything sitting on top of it had drifted a similar distance, so
the upgrades are coupled: Bootstrap, its markup across the templates, the
DataTables and toggle integrations that theme against it, and the icon font are
one change rather than several.

**The goal was that netdisco looks the same afterwards.** Visual change is
incidental, never intentional, and where the new stack could not reproduce the
old appearance that is called out below.

## What changed

| library | from | to |
|---|---|---|
| Bootstrap | 2.3.1 | 5.3.8 |
| Font Awesome | 3.2.1 | 7.3.1 |
| DataTables | 1.10.2 | 2.3.8 |
| jQuery | 1.12.4 | 3.7.1 |
| jQuery UI | 1.10.3 | 1.14.2 |
| jsTree | 3.3.12 | 3.3.17 |
| moment | 2.29.2 | 2.30.1 |
| daterangepicker | 1.2 | 3.1.0 |
| toastr | 1.1.1 | 2.1.4 |
| floatThead | 1.2.7 | 2.2.5 |
| bootstrap2-toggle | | replaced by bootstrap5-toggle 5.4.1 |
| qTip2 | 2.0.1 | removed, replaced by a Bootstrap popover |
| jquery.cookie | 1.4.1 | removed |
| underscore | 1.13.1 | removed |

d3 and Swagger UI are deliberately not here. Each is a rewrite rather than a
swap and is better reviewed on its own.

**qTip2** had one call site, the job queue log popup on hover. Bootstrap 5 ships
popovers that do the same job, so that call site moved and the dependency went.
The behavior is matched rather than approximated: qTip2 was bound on
`mouseover` with `delay: 100` and `at: 'bottom center'`, and the popover is
`trigger: 'hover'`, `delay: { show: 100 }`, `placement: 'bottom'`. Content stays
escaped, since the popover is given a string with Bootstrap's default
`html: false`.

The one behavior worth asking about is whether the log text can be selected, and
both were driven to find out rather than reasoned about: each appears on hover,
and **neither survives the pointer moving onto it**, so the text is read in
place and was not selectable before this change either.

**underscore** was loaded on every page and called by nothing: no `_.` or `_(`
anywhere in the templates, `lib/`, or netdisco's own scripts. Its one plausible
consumer was floatThead, and 2.2.5 carries its own shim with no reference to the
global at all.

### floatThead was going to be separate, and could not be

It was meant to follow on its own, because 2.x renamed both options netdisco
passes (`scrollingTop` to `top`, `useAbsolutePositioning` to `position`) and
ignores the old names without warning, so all four call sites have to move with
the file or the floating header loses its 40px offset silently.

It is here because **1.2.7 and Bootstrap 5 together destroy the job queue table
in WebKit.** Not a misalignment: one header cell spanning the full width, every
data cell empty, buttons spilling across the sidebar. The same page is correct
with 2.2.5, and correct on the current release with 1.2.7, so this pull request
introduces it. Shipping without the upgrade would have shipped a broken admin
page to Safari.

That table is the only place floatThead runs, and only when the queue has
entries, which is why it went unseen.

### A second engine-specific fix, same page

The five job queue filter fields each sit in an input group with a caret button.
Bootstrap 5 lets such a group wrap, and the field was pinned to a width it could
not give up, so field plus button filled the row exactly. Firefox and WebKit
measure that button a fraction of a pixel wider than Chromium and dropped it
onto its own line. The field may now shrink and the group no longer wraps.

### Static asset URLs now carry the release version

Assets are served with a one day lifetime and no version in the URL. That was
harmless while netdisco's markup and stylesheet changed in step. **This change
breaks them apart, so a returning visitor renders the new markup with the
previous release's stylesheet until the day is up.** Hit in practice while
testing this branch, not predicted: the page looked broken with nothing in the
console to say why.

`Web.pm`'s general template-token hook, the same `before_template` that sets
`uri_base`, now also sets `asset_version`, and the layout appends it, so
`css/netdisco.css` becomes `css/netdisco.css?v=2.101000`. Within a release
nothing changes and the one day lifetime still applies, so it costs no extra
requests in normal use.

**This covers the stylesheets and scripts the layout loads, and nothing else.**
Checked on a booted instance, those all render with the suffix, still serve
their files unchanged, and still carry `Cache-Control: max-age=86400`. The
assets that do not get a suffix are still safe, for reasons worth stating rather
than assuming. A browser may well keep any of these in cache; the point is that
none of them can be served where the new page expects something else:

| asset | versioned | why a cached copy is not a problem |
|---|---|---|
| layout css and js | yes | the URL changes each release, so the new one is always a miss |
| Font Awesome webfonts | no | the URLs changed outright, from `../font/fontawesome-webfont.woff?v=3.2.1` to `../webfonts/fa-solid-900.woff2`, so nothing new requests the old files |
| `share/public/images/` | no | no changed image URL is referenced by the new markup or styles; the six deleted images are no longer requested |
| favicon | no | unchanged, so a cached copy is the current one |
| Swagger UI | no | not touched by this change |

The snapshot fixtures render `?v=` with no value, because the fixture renderer
supplies none of the tokens the hook sets, the same reason `uri_base` is blank
there. Leaving it blank keeps those files stable across releases.

## Browser support, which this change sets for the first time

netdisco has never stated a supported-browser baseline. This change effectively
creates one, because **every library here has dropped browsers its old version
explicitly supported.** Their own words:

| library | vendor statement |
|---|---|
| Bootstrap 5.3 | "Internet Explorer is not supported. If you require Internet Explorer support, please use Bootstrap v4." Its browserslist reads `not Explorer <= 11`, `Chrome >= 60`, `Firefox >= 60`, `Safari >= 12`, `iOS >= 12` |
| DataTables 2 | "DataTables 2 and newer, plus the other software in the DataTables suite offer no support for Internet Explorer of any version." 1.x kept it |
| Font Awesome 6+ | no longer supports any version of IE; all IE-specific fallbacks and assets were removed at 6 |
| jQuery 3 | 1.x is the version to use if you need IE 6-8, Opera 12.1x or Safari 5.1+; 3.x supports IE 9-10 and Edge Legacy |

Sources: Bootstrap's [browsers and devices](https://getbootstrap.com/docs/5.3/getting-started/browsers-devices/),
DataTables' [compatibility](https://datatables.net/download/compatibility),
Font Awesome's [browser support](https://fontawesome.com/docs/web/dig-deeper/browser-support),
and jQuery's [browser support](https://jquery.com/browser-support/).

**The same story is visible in this diff, without taking anyone's word for it.**
The current tree carries **29 `progid:DXImageTransform.Microsoft.gradient`
filters** in `bootstrap.min.css`, an `<!--[if lt IE 9]>` conditional comment in
`layouts/main.tt`, and Font Awesome's `?#iefix` EOT fallback. After this change
there are none of any of them.

So the practical floor becomes recent Chrome, Firefox, Safari and Edge, and
**Internet Explorer and Edge Legacy are out.** That is not something this pull
request can decide on the project's behalf: if netdisco has deployments on
browsers that old, this is the release that stops serving them, and that call
belongs to the maintainers rather than to this change.

## How this was verified

Reviewing 159 commits of markup change by eye is not a reasonable ask, so the
appearance was checked mechanically.

A separate Playwright harness boots two checkouts, master and this branch,
against one demo database and compares them. 18 routes captured, five
interaction behaviors asserted (navbar dropdown, collapse, typeahead, tooltip,
modal) on both. Every remaining pixel difference was triaged and either fixed or
recorded below as a deviation.

**That harness is not in this pull request and not in this repository**, since
its job is to compare two checkouts of netdisco. It now lives at
https://github.com/netdisco/netdisco-visual-harness, so the evidence behind the
appearance claims here can be run rather than taken on trust. It is mentioned
because it is the evidence behind the claim that the appearance did not
materially change.

Because a pixel comparison has blind spots, two other checks were added:

- **A computed-style comparison**, element by element, across all 18 routes.
  Image comparison treats two colors as the same unless they differ by roughly
  a fifth of the full black to white range, so a large area can change shade and
  still report identical. That is how a sidebar panel losing its gray fill went
  unnoticed until the styles themselves were compared.
- **`xt/40-template-snapshots.t`**, which renders every template against a
  committed fixture. `xt/bin/regenerate-snapshots` rewrites them, so a markup
  change is reviewable as a diff of emitted HTML.

The author tests run from the tree:

```
perl -Ilib -Ixt/lib -MTest::Harness -e 'runtests(@ARGV)' xt/*.t
```

`Result: PASS`, 8 files, 758 tests. **What that does and does not cover:** the
snapshot test renders templates directly with synthetic stashes, outside Dancer
and outside a request, so it checks emitted markup only. It does not exercise
routed-page composition, and it does not exercise the release-version behavior
above, since `asset_version` is one of the tokens it has no way to set.
Appearance is the harness's job, not this test's.

### Checked in three engines, because one was not enough

The harness runs Chromium, which is the right primary target and also a real
limit: a layout that fills its container with nothing to spare can break in
another engine while every route still passes. Both defects above are exactly
that, and both were found by a person using Firefox rather than by any check
here.

So the 18 routes were also swept in Firefox and WebKit, at two viewport widths,
against both checkouts. Not by comparing images, since the engines disagree
about font rendering everywhere and a real defect would drown in that noise. The
sweep compares discrete facts that change only when a layout has actually
broken: how many rows a flex container's children occupy, and whether an element
overflows its own box.

After the two fixes that sweep is clean: 2,690 elements across the 18 routes,
with no difference between engines that the current release does not already
have. It ends by pointing the same comparison at a tree with the fixes removed,
where it reports both defects again, since a check that finds nothing proves
nothing until it has been shown to fire.

WebKit on Linux is Safari's layout engine but takes its text metrics from the
Linux font stack, so it speaks to layout behavior rather than to Safari's exact
typography.

### Bootstrap 5's palette is adopted, not repainted

The largest single difference, and deliberate. **Where a color carries meaning,
netdisco now uses Bootstrap 5's value rather than Bootstrap 2's.**

| instances | from | to | role |
|---|---|---|---|
| 322 | `#0088cc` | `#0d6efd` | link and primary |
| 270 | `#999999` | `rgba(33,37,41,.75)` | muted text |
| 115 | `#b94a48` | `#dc3545` | error |
| 90 | `#468847` | `#198754` | success |
| 80 | `#3a87ad` | `#0dcaf0` | info |
| 55 | `#c09853` | `#ffc107` | warning |

Links, alerts, validation states and badges are recognizably Bootstrap 5 rather
than the older palette. Overriding them would mean restating the framework's
semantic palette in `netdisco.css` and reconciling it by hand at every Bootstrap
release.

**The neutrals were treated the opposite way and restored.** The grays that
define netdisco's own furniture rather than the framework's identity are back:
form control text and borders (`#555555` and `#cccccc`), and the sidebar panel's
fill, border and inset shadow.

Adopt the palette, reproduce the neutrals. That is the rule the whole change
follows.

### Controls libraries draw for themselves

DataTables draws its length menu, sort arrows and pagination differently, and
the date range widgets differ by a few pixels on two report sidebars.
Re-creating these means carrying a copy of each library's styling forever.

### One change that is an improvement rather than a match

**The navbar is a single color at any window width.** The area either side of
the fixed-width container used to be a gradient that met the artwork's black
only at mid height, so a wide window showed two different blacks.

### One change that is a fix rather than a deviation

jQuery UI used to insert its ARIA live region inside the `.input-group` on the
topology page, making the real input not the first child, so Bootstrap squared
off its left corner. The current release puts that node on `body` and the input
gets its rounded corner back.

## Things worth knowing before you read the diff

**`package.json` is a manifest, not a build step.** No bundler, no minifier, no
`scripts`. It exists so dependency scanners and Dependabot can see what is
shipped under `share/public`, which previously meant opening each file and
reading its banner. Private, never published.

**`share/public/javascripts/bootstrap.min.js` is Bootstrap's bundle build**, so
Popper is inlined. The bundle emits only Bootstrap's own banner, so a short
Popper copyright notice was added above the code; the file is otherwise
byte-identical to the upstream release. Popper is in `package.json` for the same
reason.

**`netdisco.css` carries a lot of comments, deliberately.** Much of this change
reproduces behavior Bootstrap 2 gave implicitly and Bootstrap 5 does not: a
border color, a float, a checked state, a box model. Without a note saying which
rule reproduces what, they look like arbitrary overrides and will reasonably be
deleted. The cost is about 22KB gzipped, which points at something worth doing
separately: **every asset here ships unminified**, including some 1.3MB of
JavaScript, so an asset pipeline would recover far more than trimming one
stylesheet. Out of scope here.

**`xt/snapshots/` adds 120 fixture files, about 628KB, kept out of the
distribution.** `MANIFEST.SKIP` excludes them, and the test skips itself when
the tree is absent rather than reporting 120 missing files. An individual
fixture missing from a checkout is still a failure, since that means a template
was added without regenerating.

## Pre-existing issues left alone

- `share/public/swagger-ui/` has no `LICENSE` or `NOTICE`. Swagger UI is
  Apache-2.0 and section 4(d) asks for the notice to travel with
  redistributions. Better fixed alongside the Swagger UI upgrade.
- `share/public/javascripts/d3-3.5.17.min.js` contains d3 3.5.16. The filename
  has been wrong for some time; `package.json` records the real version.

## Why jQuery 3.7.1 and not 4.0

jQuery 4.0 is blocked by what sits on top of it. jsTree 3.3.17 declares
`"jquery": "^3.5.0"` as a runtime dependency, which excludes 4.x, and it has not
released since 2024. netdisco's own code calls `$.trim` nine times across five
files (`netdisco.js`, `netdisco_portcontrol.js`, `js/common.js`, `js/device.js`,
`ajax/admintask/users.tt`), and jQuery 4 removed it. Moving to 4.0 therefore
means replacing jsTree and reworking those call sites: a larger change with a
different risk profile that belongs on its own.

`netdisco.js:315` also still calls `.delegate()`, which jQuery 3 deprecates and
jQuery 4 removes. To be precise about its status: it is **deprecated and
retained**, not verified against jQuery 4. It was left alone deliberately so
that any regression from this upgrade stays attributable to the upgrade itself.


jeneric:
The Playwright harness referenced in the description is now published at
https://github.com/netdisco/netdisco-visual-harness.

I have updated the description accordingly: it previously said the harness was
yet to be offered to the project. Nothing about this pull request's contents has
changed.

ollyg:
in general there is a lot of mouseover underline I dislike, on the menubar and menu items, and so on. would prefer cell background color change if poss.

noting some glitches to address:

icon placement:
image
image
mouseover underline style:
image
image
just looks a bit weird:
image
inappropriately large text!
image
vlans tab first col width:
image
interface on addresses tab should not be centered:
image
icons could be smaller, bin icon misaligned, tickboxes misaligned:
image
after clicking bin icon the tooltip gets sticky:
image
search box too low:
image

jeneric:
All ten are addressed on the branch, each with a harness test written before the fix.

- **Mouseover underline** on the navbar, menus and tab strips: gone. Menu items fill blue on hover and tab strips fill grey, as before.
- **Submenus in the navbar menus**: open on hover and close on leave again, no click.
- **Icon placement in the Snapshot dropdown**: items padded, icons aligned with their labels.
- **Inappropriately large text**: sidebar section headings back to 13px.
- **VLANs tab first column**: columns detected as numeric or date are no longer right-aligned, and their sort arrows are back on the right.
- **Interface column on the Addresses tab**: no longer centered.
- **Icons too large, bin icon misaligned**: sidebar filter icons drawn as outlines at their old size; the bin icon is centered on its field.
- **Tickboxes misaligned**: checkbox boxes sized and spaced as before, and the seam between box and caption is square again.
- **Sticky tooltip after the bin icon is clicked**: a tooltip now hides when the pointer leaves, even while the field still holds focus.
- **Search box sitting too low**: the SNMP search row is back on one line with the gap beneath it restored.

Found while checking those, and also fixed:

- Navbar brand colour and its hover response.
- Tooltip type back to 11px; at 14px the longer labels wrapped to two lines.
- Netmap toggles printed "Off" where the markup asks for no word, and the lit word sat 4px right of where it was.
- The navbar item for the page you are on had lost its darker fill.
- Paginator: the current page was a solid blue block, and Previous on the first page was filled darker than the pages you can click.

Not changed, deliberately:

- **SNMP panel collapse at narrow widths.** The old layout is a thirteen-column row in a twelve-column grid, so its collapse is an overflow wrap and flips between stacked and side-by-side as the window narrows. The new one stacks once, at 768px.
- **Keyboard access to the navbar submenus.** They are pointer-only, as they were before. Making them keyboard-operable is an enhancement rather than a regression, so it is not in this pull request.
- Related to that, and pre-existing rather than new: the submenu category links are a plain `<a href="#">` on both versions, carrying no `role` or `aria-expanded`. They can be focused but not opened from the keyboard, and the reports inside them cannot be tabbed to. Noting it because it is a gap, not because this changes it.

Visual baselines re-accepted; harness suite green.

ollyg:
looks great! only one nit:
would it be possible to have the Find Anything text white when does not have focus? It sits like this normally:
image
but when has focus is like this:
image

making the text white when it does not have focus, so it shows up, that would be good

ollyg:
I also think the suggestions drop down from the Find Anything box should be a few pixels lower, because the first item in the suggestions is chopped off at the top by the navbar

jeneric:
Both fixed on the branch, each with a harness test written before the fix.

- **"Find Anything" invisible until the box has focus**: the prompt is light again while the field is dark, as it was. Only at rest, because the old stylesheet used one color for both states and so left it pale gray on the white focused field.
- **First suggestion chopped off by the navbar**: the list drops clear of the bar. Its first row sat 3px inside the bar and the menu's top border was painted over; both are back to what the old widget did. Checked in Chromium, Firefox and WebKit.

A third, found while checking those and fixed too: the list had stopped picking out its first row, which the typeahead did before the move to jQuery UI. That is more than looks, since it is the row Enter takes; without it, Enter submitted what you had typed. Both search boxes now open with the first row in the old blue, and Enter takes it.

jeneric:
Trimmed the migration commentary out of the shipped CSS and JS in 82b8dce3: netdisco.css is down from 99.8KB to 65.7KB and no rule changed, verified by stripping all comments from both versions and diffing what was left.

PR #1636, opened 2026-08-17
Guard against SNMP wiping ifAlias/Oper/Admin state on failed walk

Bierchermuesli:
Some devices occasionally return a whole SNMP table that's uniformly wrong for every port, without erroring — an empty ifAlias table, or every single port suddenly reporting admin-down or oper-down. Since `Discover::Properties` deletes and repopulates a device's entire port set on every discover, one bad poll silently overwrites good, previously-discovered data with garbage across every port at once. A device that's live enough to answer this SNMP session but reports every port down is a walk glitch, not reality — a truly dead device wouldn't be answering SNMP at all. This creates a bunch of unnecessary DB updates (and downstream hook/change noise), even though the next discover is usually fine again.

This PR adds a simple plausibility guard, generalized across `name` (ifAlias), `up_admin`, and `up`: it compares how many ports previously had a "good" value for each of these against how many the fresh walk returned. If the device previously had some and this walk returned none at all, it treats that as a failed/partial walk and carries the last-known values forward instead of overwriting them.

While testing this I found a second, related variant: on some devices the ifAlias walk doesn't come back blank, it comes back as the port's own short/long interface name (e.g. ifAlias `Gi3/0/14` on port `GigabitEthernet3/0/14`), even though the description is genuinely configured on the device (confirmed via `show run`). This turned out to be `i_name()` — the SNMP::Info accessor `Discover::Properties` was reading for the Name column — falling back to ifName/ifDescr on some platforms when its own ifAlias lookup comes back empty. Reading `i_alias()` directly instead (the raw ifAlias, no such fallback) removes that failure class at the source rather than needing a heuristic to detect and reverse it.

I see this daily, on random devices and at random times — mainly Cisco IOS-XE (3850, 9300, ISR1111) and Fortigate, maybe 1% of the fleet. One odd data point: when it fires, it's always either the ifAlias table or the admin/oper-status table failing — never both in the same poll. That suggests independent per-OID-subtree walk failures rather than one broader session-level failure, but I don't have a root cause yet. If anyone else has seen this and has ideas on how to trace it down further, I'd appreciate the input.

### DIFF
--- a/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
+++ b/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
@@ -420,6 +420,31 @@ register_worker({ phase => 'early', driver => 'snmp',
       };
   }
 
+  # plausibility check: SNMP occasionally returns an empty ifAlias/i_name
+  # table without erroring (seen on Cisco and Fortigate). If every port
+  # previously had a name and this poll got none at all, distrust the walk
+  # and keep the last-known names rather than wipe them all in one go.
+  {
+      my %old_names = map {($_->{'port'} => $_->{'name'})}
+        $device->ports->search(undef, {columns => [qw/port name/]})->hri->all;
+
+      my $old_named = scalar grep {defined and length} values %old_names;
+      my $new_named = scalar grep {defined $_->{'name'} and length $_->{'name'}}
+                             values %deviceports;
+
+      if ($old_named and not $new_named) {
+          warning sprintf
+            ' [%s] interfaces - i_name/ifAlias walk came back empty for all ports'
+            .' but %d had one previously - assuming failed SNMP walk, keeping existing names',
+            $device->ip, $old_named;
+
+          foreach my $port (keys %deviceports) {
+              $deviceports{$port}->{name} = $old_names{$port}
+                if exists $old_names{$port};
+          }
+      }
+  }
+
   if (scalar @{ setting('ignore_deviceports') }) {
     my $port_map = {};
 
--- a/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
+++ b/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
@@ -427,9 +427,26 @@ register_worker({ phase => 'early', driver => 'snmp',
   # reports every single port down is a walk glitch, not reality - a truly
   # dead device wouldn't be answering SNMP at all. Distrust the walk and
   # keep the last-known values rather than wipe/flap them all in one go.
+  # some platforms fall back to reporting the interface's own short or long
+  # name as i_name when the real ifAlias walk comes back empty (e.g. ifAlias
+  # "Gi3/0/14" on port GigabitEthernet3/0/14) - a genuine human-authored
+  # description doesn't echo its own port id, so don't count that as "good"
+  # ifAlias data either. Compare trailing slot/subslot/port digits rather
+  # than the letter prefix, since that's exact and vendor-abbreviation-proof.
+  my $looks_like_own_port_id = sub {
+      my ($name, $port) = @_;
+      return false unless defined $name and defined $port and length $name;
+      my ($nsuffix) = $name =~ m{(\d+(?:/\d+)*)$};
+      my ($psuffix) = $port =~ m{(\d+(?:/\d+)*)$};
+      return false unless defined $nsuffix and defined $psuffix;
+      return ($nsuffix eq $psuffix and $name =~ /^[A-Za-z]/) ? true : false;
+  };
+
   foreach my $check (
     { field => 'name',     label => 'i_name/ifAlias',
-      good  => sub { defined $_[0] and length $_[0] } },
+      good  => sub { my ($val, $port) = @_;
+                      defined $val and length $val
+                      and not $looks_like_own_port_id->($val, $port) } },
     { field => 'up_admin', label => 'i_up_admin',
       good  => sub { defined $_[0] and $_[0] eq 'up' } },
     { field => 'up',       label => 'i_up',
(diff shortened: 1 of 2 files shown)


ollyg:
Hi @Bierchermuesli thanks for raising this issue. I have two comments.

First, I don't think it's a good idea to change from i_name to i_alias because the "fallback" should be handled properly by the device class inside SNMP::Info. Netdisco has used i_name for a very long time and I'd be highly reluctant to change it. ifName, ifDescr, and ifAlias are implemented very inconsistently across vendors and platforms, see [this archived blog post about it](https://web.archive.org/web/20180829060013/http://www.irobert.info/display/IRINFO/Understanding+ifDescr,+ifName+and+ifAlias).

Second, we did see this intermittent response issue before, only on leaf fields like the serial number. So we have the SNMP [field_protection](https://github.com/netdisco/netdisco/wiki/Configuration#field_protection) feature for that. Could we look to merge your table check with this leaf check and use the same config?

Also note that the response of field_protection to unwanted results is to abort the whole job, rather than continuing with old values. [Here is the code](https://github.com/netdisco/netdisco/blob/master/lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm#L153).

Also note that field_protection checks only what Netdisco is about to put into the database, after any munging/fixing of the data, which is probably better than checking what SNMP::Info returns (i_name, etc), because there may well be ignored interfaces and so on.

I don't quite know how to make field_protection work for tables (lists like i_name) -- otherwise I would have implemented it! -- but there probably is a way. Just needs some good thought. I'd rather have a clean implementation covering both leaf and table cases if we possibly can.

ollyg:
Tagging @JeroenvIS for any comment on the i_name to i_alias change.

Bierchermuesli:
Thanks for your feedback! just to said beforehand, it was never the idea to merge as it is, I used that code to verify on the wohle fleet and spin up the discussion :) 

`field_protection` is perfect,  I hadn't actually noticed that mechanism existed before. I've implemented it: added a `device_port` category to `field_protection` (same `{field => ACL}` shape as `device`/`new_device`), scoped per-field via `acl_matches_only` exactly like the leaf checks. Also moved the check to run *after* `ignore_deviceports` filtering, so it only judges ports that are actually about to be stored (per your point about field_protection checking post-munging data) — that wasn't true before, it is now.

One thing I want to ask rather than assume: I defaulted `field_protection.device_port` to `group:__ANY__` for `name`/`up_admin`/`up` (i.e. on for everyone), matching how `device`/`new_device` already default `uptime`/`serial` on, or what you think?

On abort vs. continue for table fields — I looked into making this configurable and I don't think it should be, for a structural reason rather than just preference. `field_protection`'s leaf check runs in the *first* worker, before that worker's own `update_or_insert` commits the device row — so cancelling there is genuinely clean, nothing's been written yet. The port-table check necessarily runs later (in "get port details"), by which point the device row has *already* been committed in an earlier, separate transaction. `$job->cancel` at that point can't undo that — offering "abort" as an option here would silently not deliver the guarantee its name implies; you'd get a half-applied job (device row updated, ports stale) instead of a clean no-op.

There's also a finer reason continue is actually the *better* fit, not just the safer one: our check is field-specific (name/up_admin/up), not port-specific. If we aborted the whole ports update instead, we'd throw away every other genuinely-good field for that port too (speed, vlan, duplex, mtu, etc.) just because one field looked implausible. Reverting only the specific bad field per port while keeping the rest of that poll's fresh data is more surgical than any abort variant could be at this position in the pipeline. So I'd rather not make it configurable — happy to hear if you see a mechanism I'm missing, but I think "continue" is the structurally correct choice here, not just a default.

Bierchermuesli:
Good idea on unifying with `field_protection` — hadn't actually noticed that mechanism existed before you pointed it out, thanks. I've implemented it: added a `device_port` category to `field_protection` (same `{field => ACL}` shape as `device`/`new_device`), scoped per-field via `acl_matches_only` exactly like the leaf checks. Also moved the check to run *after* `ignore_deviceports` filtering, so it only judges ports that are actually about to be stored (per your point about field_protection checking post-munging data) — that wasn't true before, it is now.

One thing I want to ask rather than assume: I defaulted `field_protection.device_port` to `group:__ANY__` for `name`/`up_admin`/`up` (i.e. on for everyone), matching how `device`/`new_device` already default `uptime`/`serial` on. My reasoning: this prevents writing implausible data rather than suppressing existing behavior, so there's no "prior behavior" worth preserving by leaving it off by default — unlike, say, a notification/hook feature where defaulting to "on" would silently change what already happens for every existing install. But it's your call whether a bugfix like this should ship enabled by default or opt-in. Happy to flip it either way.

On abort vs. continue for table fields — I looked into making this configurable and I don't think it should be, for a structural reason rather than just preference. `field_protection`'s leaf check runs in the *first* worker, before that worker's own `update_or_insert` commits the device row — so cancelling there is genuinely clean, nothing's been written yet. The port-table check necessarily runs later (in "get port details"), by which point the device row has *already* been committed in an earlier, separate transaction. `$job->cancel` at that point can't undo that — offering "abort" as an option here would silently not deliver the guarantee its name implies; you'd get a half-applied job (device row updated, ports stale) instead of a clean no-op.

There's also a finer reason continue is actually the *better* fit, not just the safer one: our check is field-specific (name/up_admin/up), not port-specific. If we aborted the whole ports update instead, we'd throw away every other genuinely-good field for that port too (speed, vlan, duplex, mtu, etc.) just because one field looked implausible. Reverting only the specific bad field per port while keeping the rest of that poll's fresh data is more surgical than any abort variant could be at this position in the pipeline. So I'd rather not make it configurable — happy to hear if you see a mechanism I'm missing, but I think "continue" is the structurally correct choice here, not just a default.

ollyg:
looks great @Bierchermuesli I will merge, many thanks!!

To answer your question about defaulting to `group:__ANY__` - I don't see any problem with this. For new devices, old_good will be false so the "old_good and not new_good" test will never pass so it's basically a no-op. Then for discovery on existing devices it will work properly. I don't see any reason to add two-level filtering to allow per-port filtering as well as per-device, unless a user pops up and gives us a use case for that, mainly because your implementation doesn't compare the quantity of good/bad ports, only the presence of good/bad ports overall ("some previously good and none currently good").

One corner case which would trip it up would be a device with previously enabled ports which now has all ports disabled, but I figure this is highly unlikely (how would netdisco poll - unless polling on an ignored interface - quite exceptional!).

Issue #1479, opened 2026-01-08
error request to GET /ajax/content/device/ports crashed - unblessed reference


Mizaring:
## Expected Behavior
Should return the port list

## Current Behavior
Return : Search failed! Please contact your site administrator (server error).

In netdisco-web.log we get :

 error request to GET /ajax/content/device/ports crashed: Can't call method "type" on unblessed reference at /home/netdisco/perl5/lib/perl5/App/Netdisco/Web/Plugin/Device/Ports.pm line 284.

## Steps to Reproduce (for bugs)
Happens with all our Juniper EX4650-48Y-8C and QFX5120-32C, version 23.4R2-S4.11

ollyg:
Hi @Mizaring sorry about that. I've implemented a fix for the next release.

ollyg:
Hi, the commit to test is [e7b4bec](https://github.com/netdisco/netdisco/commit/e7b4becceeee3aed109ec50bf1fcde67db0672c2) or release 2.097001 which will be live in a couple of hours.

PR #1640, opened 2026-08-19
Stop the Ports tab re-reading every row once per subinterface parent

jeneric:
The device Ports tab groups subinterfaces under their parent port. To do that it
needs, for each parent name it has seen, the row carrying that name. It resolved
each one by searching the whole result list, so it re-read every row once per
parent.

On an access switch that is invisible. On a wireless controller it is the entire
request. Every port is named for an access point radio,
`70:3a:0e:cd:65:16.1`, which matches the default `subinterfaces_match` of
`(.+)\.\d+`, so nearly every port is also a parent name. The bare
`70:3a:0e:cd:65:16` port does not exist, the case #1479 describes, so every one
of those searches reads the full list and finds nothing.

This indexes the rows by port name once and looks both parents up in that index.

Raised by @rc9000 in netdisco/snmp-info#572 while adding the Catalyst 9800
wireless controller class: a ports page of 12 MB and roughly 15 s on a
controller with about 7300 ports
([measurement](https://github.com/netdisco/snmp-info/pull/572#issuecomment-5319702188)).

Measured here on a 7300-port wireless controller:

| | before | after |
|---|---|---|
| port column only | 12.98s | 1.11s |
| full default view | 11.87s | 1.54s |

Output is byte-identical in both cases.

Verified against real subinterfaces, since the controller above only exercises
the no-parent path: 120 devices that have genuinely parented subinterfaces, in
all three folding modes, 360 comparisons, all byte-identical. Across that sweep
the patched build rendered 3228 parent collapse toggles and 2882 dot-zero
collapses, neither of which can appear unless the lookups succeeded. A custom
`subinterfaces_match` was tested too, including a pattern whose captured parent
cannot line up with the hardcoded `.0`, and behaviour is unchanged.

`xt/42-subinterface-grouping-scan.t` asserts that the grouping loop resolves
ports from an index rather than by searching the result list.

### DIFF
--- a/MANIFEST
+++ b/MANIFEST
@@ -693,6 +693,7 @@ xt/36-ajax-content-response.t
 xt/37-typeahead-empty-query.t
 xt/38-swagger-ui-redirect.t
 xt/40-template-snapshots.t
+xt/42-subinterface-grouping-scan.t
 xt/50-config-override.t
 xt/bin/regenerate-snapshots
 xt/js/portsort.test.js
--- a/lib/App/Netdisco/Web/Plugin/Device/Ports.pm
+++ b/lib/App/Netdisco/Web/Plugin/Device/Ports.pm
@@ -276,8 +276,16 @@ get '/ajax/content/device/ports' => require_login sub {
         }
     }
 
+    # one pass to index the rows by port name. the two lookups below used to
+    # grep @results, re-reading every row once per parent, and on a wireless
+    # controller where every port is subinterface-shaped that is the whole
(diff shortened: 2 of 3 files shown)


ollyg:
awsome 🤩 
I'm happy to merge this and do a minor release if someone else also gives the OK

ollyg:
I realised it's not a draft. Will merge.

rc9000:
Amazing, thanks @jeneric & @ollyg !

PR #1642, opened 2026-08-20
 feat/proposal: Discovery hook only on real changes

Bierchermuesli:
Right now the `discover` event hook fires on every single poll, regardless of whether anything actually changed on the device — since something like `uptime` always differs, it's effectively unconditional. To make it meaningful I propose filter list for discover hooks: `discover_hook_ignore_changes`

In our use case we trigger a hook on `discover` to keep a NetBox inventory in sync. We only want that trigger on changes that actually matter for inventory.  Operational things are to noisy and do not belong into netbox. The configurable ignore-list this PR adds maps directly onto that distinction — here's roughly what we run:

```yaml
      discover_hook_ignore_changes:
        device:
          - uptime
          - last_discover
          - num_ports #because ignore_deviceports, num_ports is evaluated twice. 
        device_port:
          - up
          - up_admin
          - lastchange
          - speed
          - duplex
```

Real-world effect: combined with #1636, this took us from roughly 10k hook fires/day down to ~100/day — which is a healthy, meaningful change rate for our fleet size, not noise. 

side note: up_admin is synced to netbox too, but its ok when we pick it up with another change (which usually happens anyway e..g. port description change)

What it does: tracks dirty device columns (with a fix for DBIC's dirty-flag latching on fields like model/vendor that get set_column'd multiple times per job) and diffs old vs new ports (normalizing boolean-column representation, ignoring whatever's configured) to decide whether to queue the discover hook at all.

One limitation I want to flag explicitly rather than gloss over: the change-detection is a single decision per job, shared by every discover hook. If someone configures multiple hooks  wanting to react to different kinds of changes, they can't currently give each one its own ignore-list — it's all-or-nothing per job. We think that's an acceptable v1. This needs a bit more refactoring to evaluate in discover process

Default: I suggest a empty list to retain non breaking behavior.  example above might go to wiki. 

Let me know what you think, especially on the per-hook question.


lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
```diff
@@ -580,6 +607,56 @@ register_worker({ phase => 'early', driver => 'snmp',
...
+    # while our freshly built %deviceports uses the literal 'true'/'false'
+    # strings - normalize both before comparing so this isn't a false positive
+    my %bool_field = map {($_ => 1)} qw/has_subinterfaces is_master/;
+    my $normalize_bool = sub {
+      my $val = shift;
+      return (defined $val and grep {$val eq $_} qw/1 t true/) ? 1 : 0;
+    };
+
+    my %existing_ports = map {($_->{'port'} => $_)}
+      $device->ports->search(undef, {columns => [qw/
```
ollyg: is there a reason for this explicit list? can it be implicit or use get_columns which I think exists?

ollyg: what I would like is something which is not fragile to future column changes in the schema

ollyg:
Hi @Bierchermuesli many thanks! I really like the feature idea. I've added one comment to the code.

Overall I would suggest that instead of making the existing discover hook enabled/disabled, I would add a new hook type (like we have discover and new_device), perhaps called `updated_device`. Can then include your example config as the sensible default -- maybe call the setting `updated_device_hook_fields`.

To add the per hook feature now or in the future, we could add a key below `updated_device_hook_fields` called `by_hook` which is a dictionary of hook names and the updated_device_hook_fields config. We'd then need to add a `tag` or `name` optional field to hook configuration/definition.

Bierchermuesli:
Thanks for the feedback, this is a much cleaner design than what I had :) 

Implemented as suggested:

- `discover` is back to firing unconditionally, no behavior change there.
- Added `updated_device` as a new event type, gated on the diff.
- Renamed the setting to `updated_device_hook_fields`, and since this is a brand-new event nobody could already be relying on, it now ships with the sample field list as the actual default rather than empty.

Left `by_hook` out of this PR for now — didn't want to grow the diff further with the `tag`/`name` field plumbing before this core piece is settled, but I like the direction and I'm happy to build it as a follow-up once this lands, if that works for you.

deployed on my site to test

lib/App/Netdisco/Worker/Plugin/Discover/Hooks.pm
```diff
@@ -33,6 +33,14 @@ register_worker({ phase => 'late' }, sub {
       $count += queue_hook('discover', $conf);
       debug sprintf ' [%s] hooks - %s queued', 'discover', $job->device;
     }
+
+    if ($conf->{'event'} eq 'updated_device' and vars->{'device_changed'}) {
+      $count += queue_hook('updated_device', $conf);
+      debug sprintf ' [%s] hooks - %s queued', 'updated_device', $job->device;
+    }
+    elsif ($conf->{'event'} eq 'updated_device') {
+      debug sprintf ' [%s] hooks - updated_device skipped (no meaningful change)', $job->device;
```
ollyg: generally we don't log things not done (see other hooks)

lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
```diff
@@ -580,6 +607,56 @@ register_worker({ phase => 'early', driver => 'snmp',
...
+  # support for updated_device Hook - detect meaningful port-level changes
+  unless (vars->{'device_changed'}) {
+    my $ignore = (setting('updated_device_hook_fields') || {})->{'device_port'} || [];
+    my %ignore_field = map {($_ => 1)} @$ignore;
+
+    # boolean columns: ->hri bypasses DBIC's column handling, so an
+    # existing row's true/false comes back as the driver's raw "1"/"0"
+    # while our freshly built %deviceports uses the literal 'true'/'false'
+    # strings - normalize both before comparing so this isn't a false positive
+    my %bool_field = map {($_ => 1)} qw/has_subinterfaces is_master/;
```
ollyg: delete this and use native DBIC type check below

lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
```diff
@@ -580,6 +607,56 @@ register_worker({ phase => 'early', driver => 'snmp',
...
+    }
+    else {
+        PORTDIFF: foreach my $port (sort keys %deviceports) {
+            my $old = $existing_ports{$port};
+            my $new = $deviceports{$port};
+            foreach my $field (sort keys %$new) {
+                next if $ignore_field{$field};
+                my $oldval = (defined $old->{$field} ? $old->{$field} : '');
+                my $newval = (defined $new->{$field} ? $new->{$field} : '');
+                if ($bool_field{$field}) {
```
ollyg: use native DBIC field type check

lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
ollyg: ah what we need to do is get the in-storage values for the fields

lib/App/Netdisco/Worker/Plugin/Discover/Properties.pm
```diff
@@ -212,6 +212,33 @@ register_worker({ phase => 'early', driver => 'snmp',
   # support for new_device Hook
   vars->{'new_device'} = 1 if not $device->in_storage;
 
+  # support for updated_device Hook - detect meaningful device-level changes
+  {
+    my $ignore = (setting('updated_device_hook_fields') || {})->{'device'} || [];
```
ollyg: I misunderstood the setting role - needs to mention these are the IGNORED fields

ollyg (approved):
thanks! changed a few things and added some tests

PR #1643, opened 2026-08-22
Sort the Ports query in SQL instead of Perl, matching portsort.js

jeneric:
**What this is.** The Ports query now returns rows already in the order
`portsort.js` produces, via a new schema 99 function `port_sortkey()`. It
replaces the `sort_port` pass in `Ports.pm`, whose result the browser overwrote
on every page load.

**What this is not.** `portsort.js` still sorts in the browser, `ports.tt` is
unchanged, and the visible order is identical before and after. Nothing is
indexed and nothing is paged. It is not faster either: on a synthetic 7300 port
device it drops a 135ms Perl sort and adds a 191ms SQL `ORDER BY`, both
invisible beside the cost of rendering the fragment.

**Why.** `sort_port` does not define an order. Over 125 names, 4 pairs
had `cmp(a,b)` and `cmp(b,a)` both answering "greater", 19 names compared
unequal to themselves, and 50 shuffles of those names sorted into 50 different
sequences. The asymmetry is fixed here. The order the tab displays now also
exists as one SQL expression, which is what an index or a `LIMIT` would need.

**Verification.**

- `xt` 19 files / 846 tests, against 17 / 832 on master, both in a clean
  checkout.
- The served fragment matches `portsort.js`'s order for six column combinations
  on a 90 port and a 7300 port device, with no collapse error under prefetch.
- Reversing the order in the helper changes the served fragment, so the
  `ORDER BY` reaches the page.
- Each assertion in `xt/12` was checked against a deliberately broken helper.

**Not verified.** `netdisco-db-deploy` has not been run against the migration.
`xt/11` reproduces the upgrade reader's parsing rules as a proxy. The guard in
`DevicePort` exists because `SchemaVersioned` stamps the version even when a
migration fails.

**Postgres floor.** The function uses `regexp_matches ... WITH ORDINALITY` in a
`LATERAL`, needing PostgreSQL 9.4. No existing file in `share/schema_versions`
uses either.


ollyg:
took me a moment to realise what the point of this is... that it enables paged results by sending to the server the requested sort from the user and the server is able to replicate that in the database. correct me if wrong.

downsides are two:
1. that there are now three places where the same port sorting algorithm is implemented, although to be fair that's in three different languages (SQL, JavaScript, Perl), and,
2. that this is the _first ever_ function we've installed in PostgreSQL; should not make data migrations difficult as the function is not used in anything like a constraint, only the order clause.

I'll test netdisco-deploy and then merge if it works.

ollyg:
(as an aside, I suppose, now we broke the barrier of stored procs, we could one day have ACLs processed SQL 😱)

PR #1660, opened 2026-08-31
update and resync DB acls after a device renumber

jijybarnes:
a device renumber updates the ip field of all records associated with the device.
With this, corresponding ACLs in the database shall be updated replacing occurrences of the old ip address with the new one.

### DIFF
--- a/lib/App/Netdisco/DB/Result/Device.pm
+++ b/lib/App/Netdisco/DB/Result/Device.pm
@@ -413,11 +413,18 @@ sub renumber {
     device => $old_ip,
   })->delete;
 
+  $schema->resultset('AccessControlList')->search({
+    rules => { '&&' => \[ 'ARRAY[?]', $ip ] }
+  })->update({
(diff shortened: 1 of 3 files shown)


ollyg:
Hi @jijybarnes this is a good enhancement but the implementation isn't quite right. I'll provide feedback below but happy to do the implementation myself as I'm working already on another related branch.

- The work to update the database should be in `App::Netdisco::DB::Result::Device::renumber()` which is called from `Util::Device::renumber_device()`
- I don't think `sync_portctl_roles()` is needed in the Worker at all. The worker ends and that sync update is lost anyway. It has no effect on other Netdisco processes running. But if you think it fixes an issue, please let me know!

jijybarnes:
Hello @ollyg,
Honestly I think I got confused when I wrote the call to `sync_portctl_roles()`  as it would be only meaningful to call from netdisco-web once the Renumber is done.

But yeah, if you feel like doing the implementing  yourself, that'd be perfect :)

ollyg:
many thanks!

PR #1673, opened 2026-09-08
replace the last vendored widgets, and drop jQuery

jeneric:
Stacked on #1672, which must merge first. Draft: dropping jQuery itself is still to come on this branch.

The last three vendored widgets are replaced by code netdisco owns, so the search suggestions, the SNMP browser and the date range boxes can be fixed and styled here rather than waiting on an upstream release.

- jQuery UI autocomplete becomes `netdisco-typeahead.js`; `jquery-ui.min.js` and its smoothness theme are deleted
- the jsTree SNMP browser becomes server-rendered `<details>` rows fetched by htmx; jstree, its proton theme and the Titillium fonts are deleted
- the date range picker becomes a preset list above two native date inputs; `daterangepicker.js`, its stylesheet and `moment.min.js` are deleted
- toastr becomes a Bootstrap toast; `toastr.js` and its stylesheet are deleted

Three fixes ride along, each its own commit: an expired session now sends the reader to the login page and back to where they were, linked addresses sort by address rather than as text, and the browser support entry names Edge.

`Changes` gains the user-visible entries, and the site-local checker gains a rule per removed file so a site with its own copies is told what to change.

Verified: `./Build test --test_files xt/` 1170 tests over 45 files, PASS. Twenty-three node suites pass, five of them new, for the typeahead, the SNMP tree, the date range and the toast.


ollyg:
my jobqueue is not showing any rows... yours too?

minor, in the SNMP browser the tree expand and collapse icons look messed up slightly:
image

please can the device ports archived data icon be yellow/amber?
image
image

the Duplicate Devices admin report has a delete button but the icon and button are the same color (maybe change the button to default color like in manual device topology admin panel and others where similar delete buttons exist):
image

this last one is unrelated to your patch but I will be cheeky and put it in, in case Claude can work out what's going on. The Neighbor Hops text is not vertically aligned with the radio button in the device neighbors map:
image

ollyg:
the empty job queue is not really empty. Seems to be a problem with `.nd_pane-settling` from `netdisco.css` -- the content is all there, just not visible! `visibility:hidden`, for some reason

jeneric:
All four fixed, one commit each.

**Job queue.** Your diagnosis was right. A results pane is kept out of sight while its table is built, so you never see the raw unstyled rows flash before DataTables settles, and it is revealed once the table stops moving. The job queue is the only pane with no spinner, because it refreshes on a timer, and the code that reveals a pane skipped it for that same reason. So it was hidden and never shown again. 035a9a99

**SNMP tree icons.** The level line was drawn over the box instead of under it, and it was swallowing the click in Chromium and WebKit. 1496510d

**Archived data icon.** Wider than you asked for, so worth a look. Three sites were uncolored and ten were `#ffc107`; all thirteen now share one class at `#c09853`. c9cfef56

**Duplicate Devices delete button.** Now `btn` with a red icon, matching manual topology and the other six admin pages. 4122bf2d

**Neighbor Hops alignment.** The label wraps its number field, which pushed the text down below the radio. 1bdf9237

Tests added and passing for each, checked in Chromium, Firefox and WebKit.

ollyg:
awesome! thank you!
I'm just getting one remaining glitch on that tricky neighbor hops radio:
image
when neighbor hops radio button is selected the green color circle isn't aligned to the radio button home

jeneric:
Two things I had not accounted for: the circle and the dot are placed by different rules, so centering the label moved the circle and left the dot behind, and the Mac's font makes that caption wrap, which the slightly wider gap I added left no room for. 11c0455a

Four more from another sweep, including a review of the browser console:

- toggle deprecation warnings, six per page load on every device tab, c35ffac3.
- every page was in Almost Standards Mode (circa ND1, early 2000's), which Bootstrap 5 is not written for, 9a7a9a2f. Rows stand a little taller, release noted.
- an uncaught error on the Admin job queue when the pointer crossed rows, a8bf4752.
- the Modules tree connector hung below the last child, fd0de431. Long standing.

Double-click on an SNMP tree label expands a node again, 6f81e0ed.

ollyg:
chef's kiss, this looks great, thank you! 🤩 🤌

PR #1683, opened 2026-09-16
stop storing wireless clients under an SSID called "unknown"

Bierchermuesli:
`Macsuck::WirelessNodes` walks `cd11_txrate`, then looks up each index in `cd11_ssid`. Those are separate SNMP walks and do not always line up — as the existing comment a few lines up already notes for the rate table. When an index is missing from the SSID walk, the worker stores the client anyway:

```perl
my $ssid = $cd11_ssid->{$idx} || 'unknown';
```

That client is not on an SSID named "unknown", it is one we could not attribute on that pass — usually associating or roaming between the two walks. Since `node_wireless` is keyed on `(mac, ssid)`, the row becomes a second permanent identity for that client alongside its real SSID row, and nothing expires `node_wireless`, so it stays for good.

On one controller here, 46% of all `node_wireless` rows are "unknown" — more than any real SSID. Of 5807 clients holding more than one row, 5805 are a real SSID plus this artefact; exactly two are genuinely on two SSIDs. Anything joining `node` to `node_wireless` on MAC, including the web UI's own Ports tab with `n_ssid`, sees its rows roughly doubled by a value carrying no information.

This skips the client when no SSID is available, so it is picked up on a pass where the walk is complete, and reports the count so the condition is visible:

```
 [10.0.0.1] macsuck - processed 412 wireless nodes (7 skipped, no SSID reported)
```

The check is on `defined`/`length` rather than truthiness, so an SSID that is genuinely `0` is now stored correctly instead of also becoming "unknown".

**One question I cannot answer from here:** is there gear that never populates `cd11_ssid` at all? On such a device this drops the signal and rate data stored today, unusable SSID or not. If that case is real this should be a setting defaulting to current behaviour instead, and I am happy to redo it that way.

Unrelated but noticed while in here: `node_wireless` has no expiry at all, unlike `node` with `expire_nodes`. Even with this fix, stale `(mac, ssid)` pairs accumulate indefinitely. Happy to open something separate if that is worth addressing.


jeneric:
**Your open question is answerable from the SNMP::Info classes, and on that plus the measurements below this looks good to merge.**

Four classes implement `cd11_ssid`. Three read it from the same table as `cd11_txrate`, so a gap between them can only be the transient misalignment you describe: Airespace (`bsnMobileStationTable`), Aerohive (`ahClient*`), and Aironet, where `cd11_ssid` unpacks the SSID from the OID index of a sibling object in `cDot11ClientTable` rather than fetching it at all.

`Layer3::Aruba` is the exception, as its own comment notes: `staTransmitRate` from `WLSX-SWITCH-MIB`, re-keyed by stripping the BSSID, against `wlanStaAccessPointESSID` from `WLSX-WLAN-MIB`. Two MIBs, two walks.

Measured against a separate database of about 191k `node_wireless` rows, restricted to classes that can produce `cd11_ssid`:

| SNMP::Info class | devices | rows | unknown | pct |
|---|---|---|---|---|
| `Layer3::Aruba` | 21 | 50,914 | 1,981 | 3.9 |
| `Layer2::Airespace` | 13 | 9,854 | 130 | 1.3 |
| `Layer2::Aironet` | 192 | 975 | 0 | 0.0 |

Aironet zero across 192 devices, which the index-derived SSID predicts. Aruba highest, which two MIBs predict. No device is all-unknown, so nothing there loses its wireless data outright. No Aerohive in that database. (`node_wireless` records no device, so these are attributed through `node` on MAC, restricted to MACs seen on a single switch.)

`unknown` is never a real SSID there either: none of 173,766 `device_port_ssid` rows over 222 distinct SSIDs.

The precedent is three lines above your change, `next unless defined $mac` with the comment "there can be more rows in txrate than other tables". Same rule, adjacent field.

Bierchermuesli:
Thanks — that is a much better answer than I could have given, and I have pushed a change on the back of it.

I checked the class survey against the SNMP::Info source and it holds: four classes implement `cd11_ssid`, three read it from the same table as `cd11_txrate`, and `Layer3::Aruba` is the two-MIB exception its own comment describes.

But there is a gap in the framing, and it turned out to matter. `cd11_txrate` is implemented by **six** classes, not four. The other two are `Layer2::NWSS2300` and `Layer2::Trapeze`, and neither implements `cd11_ssid` — they inherit only from `SNMP::Info::Layer2`, so nothing supplies it. Restricting the measurement to classes that can produce `cd11_ssid` excluded exactly the classes at risk.

`AUTOLOAD` returns undef for an unresolvable method rather than croaking:

```perl
return unless my $meth_ref = $self->can($sub_name, @_);
```

So on those two platforms `cd11_ssid` is undef, every client currently lands under "unknown" — carrying its signal and rate data — and my patch as it stood would have dropped their wireless data outright. That is the case I was asking about, and it is real rather than hypothetical.

The fix does not need a setting. The two situations are distinguishable: a walk that returned nothing at all means the class cannot report SSIDs, and there the old fallback is still the best available answer; a gap within a walk that did return rows is the transient misalignment. Pushed as a second commit — `$have_ssids` is computed once from the walk, and only the second case skips.

Both classes do hold the SSID, as `nwss2300_sta_ssid` (`ntwsClSessClientSessSsid`) and `trapeze_sta_ssid` (`trpzClSessClientSessSsid`). Wiring those up as `cd11_ssid` looks like the real fix, but it is an SNMP::Info question rather than one for this repo, and the source comments put them in `ntwsClSessClientSessionTable` while the rates come from `ntwsClSessClientSessionStatisticsTable` — two tables, so the same two-walk exposure as Aruba. I have no such hardware and could not find those MIBs to check whether the two tables share an index, so I have not proposed it.

On the wider point, your `device_port_ssid` check is the piece of evidence I was missing — that "unknown" is never a real SSID anywhere in that data is what makes the string safe to treat as a sentinel at all. And agreed on the precedent three lines up; that adjacency is what made the original `|| 'unknown'` look deliberate rather than an oversight, which is why I asked rather than just changing it.

jeneric:
You are right, and the filter was the error rather than the data. I surveyed the classes that can produce `cd11_ssid`, when the question was which classes populate `cd11_txrate` and cannot.

`$have_ssids` is the right distinction, and I agree it needs no setting: a walk that returned nothing cannot attribute anything, while a gap inside a walk that did return rows is the transient case.

Looks good to me as it stands.

